PR #7, opened 2016-12-05
Options parameter on putFileContents()

EsSpricht:
added options parameter to putFileContents(). This allows the user to set custom headers and additional parameter for later on (eg. OAuth).

Up to now i ignored the text variant of putFileContents(). But this could be enhanced as well. 
E.g for options variable:

 var options = {
                    "headers": {
                        "custom-header": "custom property",
"custom-header2": "custom property2"
                    }
                }

Feel free to contact me.

Cheers

### DIFF
--- a/source/adapter/put.js
+++ b/source/adapter/put.js
@@ -11,13 +11,14 @@ module.exports = {
             .then(responseHandlers.handleResponseCode);
     },
 
-    putFileContents: function putFileContents(url, filePath, data) {
+    putFileContents: function putFileContents(url, filePath, data, options) {
+        //Set default headers
+        options.headers["Content-Type"] = "application/octet-stream";
+        options.headers["Content-Length"] = data.length;
+
         return fetch(url + filePath, {
                 method: "PUT",
-                headers: {
-                    "Content-Type": "application/octet-stream",
-                    "Content-Length": data.length
-                },
+                headers: options.headers,
                 body: data
             })
             .then(responseHandlers.handleResponseCode);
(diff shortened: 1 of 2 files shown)


source/adapter/put.js
```diff
@@ -11,13 +11,14 @@ module.exports = {
             .then(responseHandlers.handleResponseCode);
     },
 
-    putFileContents: function putFileContents(url, filePath, data) {
+    putFileContents: function putFileContents(url, filePath, data, options) {
+        //Set default headers
+        options.headers["Content-Type"] = "application/octet-stream";
```
perry-mitchell: What if `options.headers` doesn't exist? This would throw an exception. The options object, if used, should be standardised and robust - a single helper function could be used (private) to normalise the incoming options from the user.

Also, these lines make `Content-Type` and `Content-Length` immutable, as you override them even if the user provides them. They should be optional, for instance: `options.headers["Content-Length"] = options.headers["Content-Length"] || data.length;`.

perry-mitchell:
I've added some comments, and this MR would need an update has its tests fail.

For me to merge this, I would want both types of put requests to support the options object, including tests for the new functionality (original tests should pass unchanged).

perry-mitchell:
@EsSpricht I'm closing this as there was no activity. Due to another PR which included `options`, I've made sure that it's documented and supported on the `putFileContents` method: https://github.com/perry-mitchell/webdav-client#putfilecontentsremotepath-format-data-options

EsSpricht:
Thank you so much! sorry I just couldn't find the time.

PR #8, opened 2017-01-01
Changed queryString.escape to queryString.stringify

balu-:
I think this might be a broserify related problem, but while using webdav with broswerify i ran into the problem that the created querystring object did not have a escape method but only a queryString.stringify method.
I was wondering if this change might be taken upstream or weather this would produce problems

### DIFF
--- a/source/url.js
+++ b/source/url.js
@@ -6,7 +6,7 @@ module.exports = {
         return (username && username.length > 0) ?
             url.replace(
                 /(https?:\/\/)/i,
-                "$1" + queryString.escape(username) + ":" + queryString.escape(password) + "@"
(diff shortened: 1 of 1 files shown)


balu-:
sry i was to fast, code seems not to work please ignore

perry-mitchell:
Thanks for the attempt @balu - Pity you didn't solve it. Please feel free to open another issue if you discover another problem.

I migrated (downstream) to using webpack instead of browserify, which works very well. I personally find Browserify to be a bit clunky in its polyfills but I'd happy support it if others are willing to contribute.

Thanks!

balu:
Hello,

You mentioned the wrong balu. I think the one you wanted has a "-" after
his name.

Regards,
Balagopal.

PR #11, opened 2017-01-04
Function to get contents and corresponding headers

balu-:
To be able to decide whether a file was changed since last read the response headers of the server are most helpfull (last-modified / etag).
Would it be okay to add functions that not only return the content of a file but also the headers of the request?
I added getTextContentsAndHeaders, getFileContentsAndHeaders and corresponding testcases.


perry-mitchell:
Thanks for another PR @balu-! Just a couple of points:

 * There's a lot of code duplication in the clientFactory and get files.. it's inevitable when you duplicate functions like this. I would prefer another options object, like `{ full: false }`, where `full` => `true` means return **content**, **headers** and anything else we add in the future. Sound alright? That way we could just pass in the options like in the other file (please refer to the `deepmerge` stuff in put.js).
 * `node-fetch` [has a `raw` method on headers](https://github.com/bitinn/node-fetch/blob/3e53f0c2c86078808508aa2ca11094bb45570548/lib/headers.js#L139), so you should use `res.headers.raw()` instead of using the protected property (as it's not API-supported).

perry-mitchell:
> This option would change the function signature on the fly and would make code unable to run just by beeing set even if additionally requested information is never touched.

I'm not sure I completely understand your point here. Currently the signature (in get) is `getFileContents(url, filePath)`. What I was suggesting was just adding options to the end: `getFileContents(url, filePath[, options])`. It would be completely optional with the default behaviour performing the same procedure now (with only content being returned), so as to not break existing usages.

> i think breaking the api is not a good idea
I wholeheartedly agree here.. I don't want to break any part of the API for risk of damaging downstream usages.

> `getFileContentsExtended`
I see what you're meaning here, but I think if we were to use the options as an _optional_ parameter we might not need to change existing function names.

> Don't get me wrong your the one with more experience and if you think one function this is the way to go i'll follow.
I'm flattered, but I'm still learning myself 😄  I appreciate the discussion!

> Final note : Is the module get used seperatly somewhere (in other terms may the api be changed here)? If not the code dublication can be kept very short and in clientFactory.
Nope, only internally. The duplication wouldn't be an issue if we opt for the options parameter.

balu-:
> I'm not sure I completely understand your point here. Currently the signature (in get) is getFileContents(url, filePath). What I was suggesting was just adding options to the end: getFileContents(url, filePath[, options]). It would be completely optional with the default behaviour performing the same procedure now (with only content being returned), so as to not break existing usages.

Sorry for not beeing clear. I meant the signature of the return value. 
Code that is now working with `getFileContents(url, filePath)` is expecting a Buffer (or String in getText case) as result of the returned promise. If later on someone appends the option for headers to the call the result of the returned promise will be a very different and the application can no longer work, also the option should just add information.
In my option the result should always be an `Object` with at least the `content` attribute, but this would be an api breaking change -> therefor i was thinking new function. 
(Sorry i think i could not make my thoughts as clear as i would have liked to, maybe the code helps a little)

perry-mitchell:
> If later on someone appends the option for headers to the call the result of the returned promise will be a very different and the application can no longer work

Of course, but this is to be expected. The developer would then be asking for more information, and having read the docs they should expect different output, right? Extracting only the contents from such an update would be trivial:

```
.then(function({ contents, headers } = {}) {
    // contents is the file data as you suggested,
    // headers is the headers object
});
```

_(Providing they use ES6)._

I'm almost in the mind to consider just breaking existing functionality to support the object output without the need for a flag or new functions..

perry-mitchell:
This is fast turning into just an architectural question 🤔

balu-:
> Of course, but this is to be expected. The developer would then be asking for more information, and having read the docs they should expect different output, right? 
Yeah your totaly right especialy because extraction is so simple. 

While writing the last chunk of code i got the idea that there might be a `dirty solution`.
Not saying that i like it ... just putting it out there : 
```
function getFileContents(remoteFilename, formatOrOptions) {
    if(typeof formatOrObject === 'String'){
         <result = Buffer / String>
     } else {
        <result = Object>
     }
}
```
Because if there is an options object, the `format` string should no longer bee needed.

perry-mitchell:
> Because if there is an options object, the format string should no longer bee needed.

I don't really like this option, as it's kind of the same reason I don't like the encoding option in Node's fs module for `readFile` etc. - it's just a little bit messy imo.

I've had a bit of a think about our whole conversation, and I find myself going back to your original suggestion regarding having 2 functions. What do you think about the following?

**clientFactory**

```javascript
function getFile(remoteFilename, format) { } // gets content + headers in object format

function getFileContents(remoteFilename, format) { } // returns content only
```

**adapter / get***

```javascript
function getFile(url, filePath) { } // makes raw request

function getFileContents(url, filePath) { } // calls `getFile`, outputs buffer + headers

function getTextContents(url, filePath) { } // calls `getFile`, outputs text + headers
```

So the only changes here would be:

 * Abstract all functionality in the factory to `getFile`, which returns the most verbose output
 * Make `getFileContents` call `getFile` in the factory, returning only the contents (so to not break the API)
 * Add `getFile` to the adapter, which makes the raw request and kind of looks like the following example
 * Make `getTextContent` and `getFileContent` call `getFile` in the adapter, but return the headers as well (all in an object)

**get adapter: getFile**

```javascript
function getFile(url, filePath) {
        return fetch(url + filePath).then(responseHandlers.handleResponseCode);
}
```

I feel that this would be the more succinct way of fetching data..

perry-mitchell:
Hi @balu- I see you've made some other changes. Perhaps it might be best to keep this PR on course and quite simple? There's a lot to discuss if we modify the API for all of the other methods.

balu-:
Sorry @perry-mitchell I hadn't thought about the PR while making those changes. 
Should I move them (away or in a different PR) ? (They are just additional options params for the alter functions => so non api breaking)

perry-mitchell:
Yes, it would be good as a different PR. But perhaps start an issue first so we can discuss adding options everywhere architecturally.

Thanks.

perry-mitchell:
Hi @balu- - do you have bandwidth to update this MR? I've opened #16 in case you don't.

balu-:
hi @perry-mitchell 

sorry im completly swamped at the moment (no free time).
hopefully that will change in the future but i cant predict a timeframe atm

perry-mitchell:
Since there have been major updates, and this branch carries extra out-of-scope modifications, it might be best if it were done in a new PR. I'll close this for now.

PR #15, opened 2017-01-17
Check double slash

ardencod3:
Make sure it won't add double slash that will make path.basename return nothing.


perry-mitchell:
Thanks @ardencod3

Issue #4, opened 2016-10-14
Add error codes for HTTP errors


perry-mitchell:
Add error codes for HTTP requests

perry-mitchell:
This is already handled it seems. Errors contain `status` property.

PR #26, opened 2017-04-06
add http status code to the thrown exception.

ccoenen:
easy way to add some information to the thrown exception.
More elegant would be a custom error class, but this gets the job done quite nicely, too. This is related to
https://github.com/perry-mitchell/webdav-fs/issues/38 and https://github.com/perry-mitchell/webdav-client/issues/4


perry-mitchell:
Thanks @ccoenen

perry-mitchell:
Released in 0.6.0.

Issue #47, opened 2017-06-27
Encode & decode URI


kistora:
I have an issue with your webdav-client.
### Config
- NodeJS v.8.0.0
- Webdav-client v0.10.0
- OwnCloud v10.0.2.1 as webdav server
### Description
- I want to get the content of the following directory: `/têst/`
- I use the getDirectoryMethod like: `client.getDirectoryContents("/têst");`
- But i have an Internal Server Error (500)
### Proposed fix
The problem is that the URI is not encoded before sending the request using fetch.
So I transform the call of getDirectoryMethod like:  `client.getDirectoryContents(encodeURI("/têst"));`
## Issue 2
With the fix of issue 1, another issue appears:
### Description
- I want to get the content of the following directory: `/têst/`
- I use the getDirectoryMethod like: `client.getDirectoryContents(encodeURI("/têst"));`
- But the getDirectoryMethod return an empty array even if the folder is not.

So, the issue is that the path is not encoded before fetching. But the path is directly used to match the filename in the parser.
I think that the webdav module should handle pre-encoded path, so here is the proposed fix:
### Proposed fix
You have to add this following line: `dirPath = decodeURI(dirPath);` before line 64 in the file `parse.js`.

Thank you for your work.

perry-mitchell:
Thanks @kistora. I think this could be easily added. Would be great to have a failing test added for it. I'll try to get to it soon if no on else does first.

kistora:
@perry-mitchell So I've done it. #48

PR #48, opened 2017-06-28
Decode uri implementation

kistora:
Fix the issue [#47](https://github.com/perry-mitchell/webdav-client/issues/47).


test/specs/adapter-get.spec.js
```diff
@@ -118,6 +118,13 @@ describe("adapter:get", function() {
                         });
                     });
             });
+            it("gets all objects in directory with special character", function() {
+                return getAdapter
+                    .getDirectoryContents(SERVER_URL, "/test%20test/")
```
perry-mitchell: @kistora This is great, but only tests spaces - could we add another "special" character to test? Say some characters like å and 几 ?

Also, I don't see any encode call..

kistora:
@perry-mitchell I have implemented your request. Good job btw!

perry-mitchell:
Thanks @kistora - Planning a larger release soon (1.0.0), which I'll make this a part of.

PR #51, opened 2017-07-06
add npmignore

pladaria:
We had problems in our CI pipeline because for some reason, the folder with strange characters was causing an error in Jenkins.

I know it's not your fault, but anyway, the test folder is not needed in the npm package.

Please take into consideration this PR

Thank you for your great job!


perry-mitchell:
Thanks @pladaria - I agree. I'll merge this shortly, and it'll go out as part of the next feature/bugfix release. Cheers 👍

PR #60, opened 2017-08-08
Const/Let: fix #56

### LINKED ISSUE #56: Update code style -> const/let
Use scoped declarations to neaten the library.

zh99998:


source/factory.js
```diff
@@ -1,9 +1,9 @@
-var deepmerge = require("deepmerge");
+const deepmerge = require("deepmerge");
 
-var authTools = require("./auth.js"),
+const authTools = require("./auth.js"),
     urlTools = require("./url.js");
```
perry-mitchell: These should be updated stylistically, to something like this:

```javascript
const authTools = require("./auth.js");
const urlTools = require("./url.js");
```

(no comma-separation of lines when using const)

PR #61, opened 2017-08-08
Add prettier formatting: fix #57

### LINKED ISSUE #57: Add prettier formatting
Add prettier to force code formatting.

zh99998:
notice that prettier recommends use 2 space as indent, not 4.

since Google https://github.com/angular/angular-cli/blob/master/.editorconfig#L8 、 StandardJS https://standardjs.com/ recommends 2 space too. maybe it is better?


.editorconfig
```diff
@@ -4,7 +4,7 @@ insert_final_newline = true
 
 [*.{js,json}]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
perry-mitchell: I prefer 4 spaces, and find it more readable.

source/auth.js
```diff
@@ -1,9 +1,9 @@
 "use strict";
 
 function generateBasicAuthHeader(username, password) {
-    return "Basic " + (new Buffer(username + ":" + password)).toString("base64");
+  return "Basic " + new Buffer(username + ":" + password).toString("base64");
```
perry-mitchell: I think having the braces around the `new Buffer()` is much nicer. `new Something().blah()` is ugly and misleading.

perry-mitchell: I wonder if there's a prettier setting for this..

perry-mitchell: I guess there isn't.. nevermind.

zh99998:
yes you are the boss, you always right, We Will Follow You Only

zh99998:
ohh it failed..

zh99998:
done.

new Buffer() call is deprecated in node.js now since you don't like it.

.editorconfig
```diff
@@ -7,7 +7,7 @@ indent_style = space
 indent_size = 4
 
 [package.json]
-indent_size = 2
+indent_size = 4
```
perry-mitchell: I just meant for the code, package.json + other existing configs could stay at 2 spaces for simplicity. Many programs that generate them automatically only use 2 spaces (like `npm install` for example). Best leave it like that perhaps

perry-mitchell:
> yes you are the boss, you always right, We Will Follow You Only

Perhaps you could avoid the patronising comments, if you don't mind. I think I've been fair in my requests..

> new Buffer() call is deprecated in node.js now since you don't like it.

Thanks for changing that, looks much better now.

I added only 1 more comment. Otherwise this MR looks very good.

zh99998:
fixed package.json indent size back to 2

PR #62, opened 2017-08-25
transpiler

zh99998:
request add a transpiler again.

my angular project using webdav as dependency is broken now.
there is no transpiler with npm modules in angular-cli's build process.

as
https://github.com/angular/angular-cli/issues/3634#issuecomment-269785768

seems release a pre-transpiled es5 version is best practice. and angular-cli only support that.

------------
another reason.
it allows you can using new js features like import, async functions without worry runtime engine version.
http://node.green/
node 4 only support 58% of es6 feature, many valid es6 syntax like `destructuring` `default function parameters` are not supported in it.


perry-mitchell:
I agree that a transpiled copy would be a good idea. I'll merge later and will most likely integrate webpack as well, for a more complete build process. Thanks for this.

package.json
```diff
@@ -2,9 +2,11 @@
   "name": "webdav",
   "version": "1.0.1",
   "description": "WebDAV client for NodeJS",
-  "main": "source/index.js",
+  "main": "dist/index.js",
```
perry-mitchell: I don't agree with changing the main to this though. I will most likely want the following:

 * Using in node: `require("webdav");`
 * Using in browser: `require("webdav/browser")` or `import createClient from "webdav/browser"`

I want the primary use case to remain node without transpilation. The browser transpiled copy can be it's own product.

perry-mitchell: You don't have to change this, I'll do so when I merge.

zh99998:
consider node 4.
we have meet this trouble once
https://github.com/perry-mitchell/webdav-client/pull/60#pullrequestreview-54830798
It's correct js syntax, and IDE say it's correct, lint say it's currect,
test say it's current, everyone say it's correct.

but after I commit, node 4 CI say it's wrong. do you think it's awesome ?

will you avoid using all new js features, just because node 4 don't support
it ?



2017-08-25 17:36 GMT+08:00 Perry Mitchell <notifications@github.com>:

perry-mitchell:
> will you avoid using all new js features, just because node 4 don't support it ?

That's the idea, yes. You don't always need the newest features, for the program to work well and be easily maintainable. Sure, arrow funcs and generators are fun and can make code more readable, but they're far from necessary.

The version of node I choose is from a support angle - I'm not in love with ES5, but I am comfortable using it.

image
_Taken from [here](https://nodejs.org/static/documents/casestudies/Nodejs_2017_User_Survey_Exec_Summary.pdf)._

If you don't agree with what I'm saying then I guess we're at an impasse. I do appreciate your feedback and contribs.. please don't let my resolute attitude here dissuade you.

perry-mitchell:
I may use webpack to build both node and browser versions after merging this.. Let's see. If that's the case then yeah, I'll obviously have no issue using newer ES features.

Issue #66, opened 2017-11-12
Special characters in paths break output


perry-mitchell:
There may be encoding issues or flat-out errors when referencing paths that have non-latin characters in them.

nickmisasi13:
I have forked your repo and made a fix for this as I needed it for my own purposes. Would you like me to submit a PR?

nickmisasi13:
Opened PR in: #71

PR #71, opened 2018-02-09
Allow special characters (like #)

nickmisasi13:
Currently the module does not support uploading files with a name like `Data #1.csv`. I've added functionality to parse the link and encode any special characters before it makes any requests to the Webav server.

### DIFF
--- a/source/request.js
+++ b/source/request.js
@@ -4,8 +4,37 @@ const nodeFetch = require("node-fetch");
 
 let fetchMethod = nodeFetch;
 
-function request(url, options) {
-    return fetchMethod(url, options);
+/**
+ * return a URL encoded path, including encoding for special characters like #
+ * @param {String} uriPath that we're trying to encode
+ * @returns {String} encoded uri path
+ */
+function encodeURIPath(uriPath) {
+    const uriSections = uriPath.split("/");
+    for (let i = 0; i < uriSections.length; i += 1) {
+      uriSections[i] = encodeURIComponent(uriSections[i]);
+    }
+    return uriSections.join("/");
+}
+
+/**
+ * Parse the url in a way that won't crash if the URL is invalid (conventional methods will not accept characters like the #)
+ * @param {String} fullUrl that was sent to the request function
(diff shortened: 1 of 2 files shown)


nickmisasi13:
Fixes #66

source/request.js
```diff
@@ -4,8 +4,37 @@ const nodeFetch = require("node-fetch");
...
+function splitInvalidURL(fullUrl){
+    const sections = fullUrl.split("/");
+    const url = {};
+    url.protocol = sections[0] + "//";
+    url.hostname = sections[2];
+    url.path = "/" + sections.slice(3, sections.length).join("/");
+    return url;
+}
+
+function request(fullUrl, options) {
```
perry-mitchell: I wouldn't recommend changing this file to perform this work. For one, splitting the URL like this is a bit redundant and unsafe. Ultimately only the _path_ needs encoding ([here](https://github.com/perry-mitchell/webdav-client/blob/master/source/interface/putFile.js#L27)), which could be done in the interface files like [putFile.js](https://github.com/perry-mitchell/webdav-client/blob/master/source/interface/putFile.js).

Ideally there'd be some helper method to perform the encoding like `encodeRequestPath(path)` that you could use on all get/put/mkdir requests.

perry-mitchell:
Thanks for submitting @nickmisasi13 - I've commented regarding where the work has been done. I would prefer it to be performed in the actual request methods themselves rather than the request getter. Also the tests are failing, which must pass for a review to go through.

source/request.js
nickmisasi13: I tried a bunch of different splitting methods, particularly `url = new URL(...)` however since something like 'https://webdav-server.com/webdav/file #1.json` has the `#` character, the path is only parsed up to that point (`#1.json` is removed). How would you go about parsing?

I'll look into moving the encoding around. I went this way to keep the code more DRY.

perry-mitchell:
Thanks for the work, but unfortunately I'm going to close this one in favour of #76. The new PR addresses the encoding for only the paths.

PR #83, opened 2018-03-15
Added getFileDownloadLink function

yoeven:
It returns a remote URL with an authentication query that can be directly accessed as a download link, for example in `<a href="">` or `window.location` to prompt a save dialog. Works similarly to `getFileContents`.

### DIFF
--- a/source/factory.js
+++ b/source/factory.js
@@ -151,6 +151,18 @@ function createClient(remoteURL, username, password) {
         },
 
         /**
+         * Get the contents of a remote file
+         * @param {String} remoteFilename The file link to fetch
+         * @param {OptionsHeadersAndFormat=} options Options for the request
+         * @memberof ClientInterface
+         * @returns {String} Returns a http url string
+         */
+        getFileDownloadLink: function getFileDownloadLink(remoteFilename, options) {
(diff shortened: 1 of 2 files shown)


yoeven:
I know Owncloud/Nextcloud WebDAV service will work (tested) and it should work on the other WebDAV services that you've stated in your `README.md` as it follows the same options and calls. The only issue is that it exposes the username & password in the URL, so links can't be shared with others or have to be encrypted somehow.

perry-mitchell:
Ah cool.. Well nice that it seems to be a more common feature. You're right that the URL authentication method is not ideal, but that choice can be up to the users (with fair warning of course). I'll review this a bit later, and then will most likely add something in the readme about it.

It's cool that this would allow the request to be handled externally using the URL - I'd see this as almost the best use case for this method.

perry-mitchell:
@yoeven could you please add a unit test for this new method?

yoeven:
Sorry @perry-mitchell, I've never worked with unit testing before but will look into it

perry-mitchell:
Thanks for this @yoeven - No unit test necessary for this PR, I'll write one myself now. Cheers!

PR #84, opened 2018-03-16
Add OAuth2 token authentication

Treyone:
This adds the possibility to pass an OAuth2 token to the `createClient` method. The token descriptor must comply with the RFC 6749 standard : 
```

     {
       "access_token":"2YotnFZFEjr1zCsicMWpAA",
       "token_type":"example",
       "expires_in":3600,
       "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
       "example_parameter":"example_value"
     }
```
Testing if the 1st parameter (`username`) has a `token_type` attribute seemed to be the lightest non-breaking solution. The cleanest one would have been to change the function signature to `createClient(remoteURL, credentials)`, with `credentials` being either a token or `{username:'xxx', password:'yyy'}`, but I wanted to avoir breaking changes.

The `webdav-server` npm used in the tests does not seem to implement this authentication scheme. Therefore I could test by adding a mocking tool like `nock` to the project, but this needs to be discussed first.


source/factory.js
```diff
@@ -57,6 +64,9 @@ function createClient(remoteURL, username, password) {
     if (username && username.length > 0) {
         baseOptions.headers.Authorization = authTools.generateBasicAuthHeader(username, password);
     }
+    if (username && username.token_type) {
```
perry-mitchell: Wouldn't it be perhaps better to avoid the `length` check of the first if statement? Perhaps something like:

```javascript
if (username) {
	baseOptions.headers.Authorization = typeof username === "object" ?
		authTools.generateTokenAuthHeader(username) :
		authTools.generateBasicAuthHeader(username, password);
}
```

perry-mitchell:
Thanks @Treyone! This is cool.

> Testing if the 1st parameter (username) has a token_type attribute seemed to be the lightest non-breaking solution

I agree - for now this is the best way. 2.0 will include a more robust method of authenticating.

> The webdav-server npm used in the tests does not seem to implement this authentication scheme. Therefore I could test by adding a mocking tool like nock to the project, but this needs to be discussed first.

I like this approach.. Would you be up to adding it?

Treyone:
Done, ready for review :)

source/factory.js
```diff
@@ -61,11 +61,12 @@ function createClient(remoteURL, username, password) {
         remotePath: urlTools.extractURLPath(remoteURL),
         remoteURL: remoteURL
     };
-    if (username && username.length > 0) {
-        baseOptions.headers.Authorization = authTools.generateBasicAuthHeader(username, password);
-    }
-    if (username && username.token_type) {
-        baseOptions.headers.Authorization = authTools.generateTokenAuthHeader(username);
+
+    if (username) {
```
perry-mitchell: 👍

perry-mitchell:
Good update! The tests are currently failing, however, due to the use of async/await. Would you mind returning a promise?

Treyone:
I adapted the code to pass the older node versions used in your travisCI configuration. I think everything os now ok :)

perry-mitchell:
Looks great @Treyone - Thanks! 😊

PR #86, opened 2018-03-21
Added a new function and change some es6 features to es5

yoeven:
**1.** I've added `getFileUploadLink `which works like `getFileDownloadLink`. Similarly, it also allows users to use the request module to process the upload from their end.

A simple example: 
`var link = client.getFileUploadLink(RemoteDirToUploadTo + "/" + filename);`

A simple example with the [request](https://github.com/request/request) and [process-request ](https://github.com/IndigoUnited/node-request-progress)modules:

```
    var data = fs.createReadStream(LocalDirFile);
    var options = {
        url: link,
        body: data,
    };

    var r =  request.put(options);

    progress(r, {})
        .on('error', function (err) {
            progressUpdateFunction("Error");
        })
        .on('end', function () {
            progressUpdateFunction("Done");
        });

    var amountdone= 0;
    data.on('data', function (data) {
        amountdone+= data.length;
        progressUpdateFunction((amountdone/ filesize) * 100);
    })
```

**2.** I've changed few es6 features to es5 features as react building process takes any node modules with es6 syntax as an error and breaks the build.

### DIFF
--- a/source/auth.js
+++ b/source/auth.js
@@ -4,10 +4,10 @@ function generateBasicAuthHeader(username, password) {
     return "Basic " + Buffer.from(username + ":" + password).toString("base64");
 }
 function generateTokenAuthHeader(tokenInfo) {
-    return `${tokenInfo.token_type} ${tokenInfo.access_token}`;
+    return tokenInfo.token_type + " " + tokenInfo.access_token;
 }
 
 module.exports = {
-    generateBasicAuthHeader,
-    generateTokenAuthHeader
+    generateBasicAuthHeader: generateBasicAuthHeader,
+    generateTokenAuthHeader: generateTokenAuthHeader
 };
--- a/source/auth.js
+++ b/source/auth.js
@@ -3,7 +3,11 @@
 function generateBasicAuthHeader(username, password) {
     return "Basic " + Buffer.from(username + ":" + password).toString("base64");
 }
+function generateTokenAuthHeader(tokenInfo) {
+    return `${tokenInfo.token_type} ${tokenInfo.access_token}`;
+}
 
 module.exports = {
-    generateBasicAuthHeader
+    generateBasicAuthHeader,
+    generateTokenAuthHeader
 };
--- a/source/auth.js
+++ b/source/auth.js
@@ -4,10 +4,10 @@ function generateBasicAuthHeader(username, password) {
     return "Basic " + Buffer.from(username + ":" + password).toString("base64");
 }
 function generateTokenAuthHeader(tokenInfo) {
-    return tokenInfo.token_type + " " + tokenInfo.access_token;
+    return `${tokenInfo.token_type} ${tokenInfo.access_token}`;
 }
 
 module.exports = {
-    generateBasicAuthHeader: generateBasicAuthHeader,
-    generateTokenAuthHeader: generateTokenAuthHeader
+    generateBasicAuthHeader,
+    generateTokenAuthHeader
 };
--- a/source/factory.js
+++ b/source/factory.js
@@ -212,6 +212,19 @@ function createClient(remoteURL, username, password) {
         },
 
         /**
+         * Get the upload link
+         * Only supported for Basic authentication or unauthenticated connections.
+         * @param {String} remoteFilename The path of the remote file location
+         * @param {PutOptions=} options The options for the request
+         * @memberof ClientInterface
+         * @returns {String} A upload URL
+         */
+        getFileUploadLink: function getFileUploadLink(remoteFilename, options) {
+            var putOptions = merge(baseOptions, options || {});
+            return putFile.getFileUploadLink(remoteFilename, putOptions);
+        },
+
+        /**
          * Stat a remote object
          * @param {String} remotePath The path of the item
          * @param {OptionsWithHeaders=} options Options for the request
--- a/source/factory.js
+++ b/source/factory.js
@@ -239,5 +239,5 @@ function createClient(remoteURL, username, password) {
 }
 
 module.exports = {
-    createClient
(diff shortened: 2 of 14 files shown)


source/auth.js
```diff
@@ -4,10 +4,10 @@ function generateBasicAuthHeader(username, password) {
     return "Basic " + Buffer.from(username + ":" + password).toString("base64");
 }
 function generateTokenAuthHeader(tokenInfo) {
-    return `${tokenInfo.token_type} ${tokenInfo.access_token}`;
+    return tokenInfo.token_type + " " + tokenInfo.access_token;
```
perry-mitchell: No need for this - Babel is used before publishing, so this will be transpiled. The tests pass for node 4, which is my earliest support version:

image

Anything outside of this is not currently in scope.

source/auth.js
```diff
@@ -4,10 +4,10 @@ function generateBasicAuthHeader(username, password) {
...
 }
 
 module.exports = {
-    generateBasicAuthHeader,
-    generateTokenAuthHeader
+    generateBasicAuthHeader: generateBasicAuthHeader,
+    generateTokenAuthHeader: generateTokenAuthHeader
```
perry-mitchell: This can be reverted due to it not being required.

source/factory.js
```diff
@@ -226,5 +239,5 @@ function createClient(remoteURL, username, password) {
 }
 
 module.exports = {
-    createClient
+    createClient: createClient
```
perry-mitchell: Same here..

source/interface/createStream.js
```diff
@@ -67,6 +67,6 @@ function getFileStream(filePath, options) {
 }
 
 module.exports = {
-    createReadStream,
-    createWriteStream
+    createReadStream: createReadStream,
+    createWriteStream: createWriteStream
```
perry-mitchell: Same here..

source/interface/directoryContents.js
```diff
@@ -70,5 +70,5 @@ function getDirectoryFiles(result, serverBasePath, requestPath) {
 }
 
 module.exports = {
-    getDirectoryContents
+    getDirectoryContents: getDirectoryContents
```
perry-mitchell: And here..

source/interface/getFile.js
```diff
@@ -36,7 +36,7 @@ function getFileLink(filePath, options) {
         }
         const authPart = options.headers.Authorization.replace(/^Basic /i, "").trim();
         const authContents = Buffer.from(authPart, "base64").toString("utf8");
-        fetchURL = fetchURL.replace(/^https?:\/\//, `${protocol}://${authContents}@`);
+        fetchURL = fetchURL.replace(/^https?:\/\//, protocol + "://" + authContents + "@");
```
perry-mitchell: And here 😊

source/interface/putFile.js
```diff
@@ -30,6 +30,22 @@ function putFileContents(filePath, data, options) {
...
+function getFileUploadLink(filePath, options) {
+    var fetchURL = joinURL(options.remoteURL, encodePath(filePath));
+    fetchURL += "?Content-Type=application/octet-stream";
+    var protocol = /^https:/i.test(fetchURL) ? "https" : "http";
+    if (options.headers.Authorization) {
+        if (/^Basic /i.test(options.headers.Authorization) === false) {
+            throw new Error("Failed retrieving download link: Invalid authorisation method");
+        }
+        var authPart = options.headers.Authorization.replace(/^Basic /i, "").trim();
+        var authContents = Buffer.from(authPart, "base64").toString("utf8");
```
perry-mitchell: Please use const/let (stylistic choice, to match rest of repo).

source/interface/putFile.js
```diff
@@ -30,6 +30,22 @@ function putFileContents(filePath, data, options) {
...
+        fetchURL = fetchURL.replace(/^https?:\/\//, protocol + "://" + authContents + "@");
+    }
+    return fetchURL;
+}
+
 module.exports = {
-    putFileContents: putFileContents
+    putFileContents: putFileContents,
+    getFileUploadLink: getFileUploadLink
```
perry-mitchell: Same here..

source/merge.js
```diff
@@ -5,5 +5,5 @@ function merge(...args) {
 }
 
 module.exports = {
-    merge
+    merge: merge
```
perry-mitchell: And here..

yoeven:
Hi, @perry-mitchell I've reverted the es5 changes as stated. I seem to get a minifying file error from building a react project if I use the es6 features for this module. Based on few threads that I read online, it seems like es6 is an issue even though the oldest nodejs supports it (thread links [here ](https://github.com/ccxt/ccxt/issues/299)& [here](https://github.com/facebook/create-react-app/issues/3734)).

source/interface/putFile.js
perry-mitchell: @yoeven These can still be changed..

perry-mitchell:
That sounds strange @yoeven - What version of node are you running? You can see [here](https://github.com/perry-mitchell/webdav-client/blob/master/package.json#L5) that webdav uses the built version of this library when required by other libraries. The code that is seen externally should be Node v4 compatible, so it shouldn't cause any issues. At this stage I would suggest that it's a configuration issue in your react app and not something to do with this library.

If you disagree or find an issue, definitely feel free to open a separate issue so we can debug it.

perry-mitchell:
Thanks! And thanks for the submission. I’ll merge soon.

perry-mitchell:
Just to remind: I'll merge after the latest changes are made. Thanks!

PR #88, opened 2018-03-25
Dir contents trailing slash bug

perry-mitchell:
Fix a bug where requesting paths like `/some/path/` using `getDirectoryContents` would show the parent directory in the output.


perry-mitchell:
This should address perry-mitchell/webdav-fs#56

PR #92, opened 2018-04-25
Added support for an http(s).Agent to get passed to node-fetch

Jakobud:


source/factory.js
```diff
@@ -62,11 +63,12 @@ const stats = require("./interface/stat.js");
...
  *      });
  */
-function createClient(remoteURL, username, password) {
+function createClient(remoteURL, username, password, agent) {
     const baseOptions = {
         headers: {},
         remotePath: urlTools.extractURLPath(remoteURL),
-        remoteURL: remoteURL
+        remoteURL: remoteURL,
+        agent: agent ? agent : undefined
```
perry-mitchell: This will be undefined by default, so you could just use `agent` imo..

source/factory.js
```diff
@@ -43,6 +43,7 @@ const stats = require("./interface/stat.js");
  * @param {String} remoteURL The remote address of the webdav server
  * @param {String=} username Optional username for authentication
  * @param {String=} password Optional password for authentication
+ * @param {Agent} agent Optional http(s).Agent instance, allows custom proxy, certificate etc. Gets passed to node-fetch
```
perry-mitchell: Let's use `http.Agent|https.Agent` as the param type, so that hopefully most smart editors will pick up where it's coming from. This could also be marked optional:

`@param {http.Agent|https.Agent=} ...` for instance.

Jakobud: Technically, according to the Node.js documentation it's called "Agent":

https://nodejs.org/api/http.html#http_new_agent_options

I would think that any Intellisense type setups would be able to understand that. But yeah that could be easily changed if you think it would help.

perry-mitchell: I see your point.. Just thought `Agent` was a bit ambiguous. Let's leave it as it is..

Jakobud: I think as long as the documentation and JSDoc is clear on what it is and how to use it then it would be good to go.

perry-mitchell:
Thanks @Jakobud !

PR #95, opened 2018-05-03
Fix special character issue (#93)

### LINKED ISSUE #93: Name decoding is broken for already encoded characters.
I was facing this issue today, and managed to implement a simple failing test.

If a directory has a name containing encoded character it will fail to get the correct directory name.

I have added a directory with name `two%20words`, containing a single file.
And testing `getDirectoryContents` on this folder.

Step to reproduce:
• Add `./test/serverContents/two%20words/file2.txt`
• Add the following test in `getDirectoryContents.spec.js`
   ```
      it("returns only the directory content when encoded special characters in name", function() {
        return this.client.getDirectoryContents("/two%20words").then(function(contents) {
            console.log(contents);
            expect(contents).to.have.lengthOf(1);
            expect(contents[0].basename).to.equal("file2.txt");
        });
    });
   ```
• npm t

silto:
Fixes path decoding on file paths like "text%20folder" by removing useless decodings.
Fixes #93 (in our cases, on pCloud and NextCloud webdav interfaces)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58,7 +58,7 @@
     "@sinonjs/formatio": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
-      "integrity": "sha512-ls6CAMA6/5gG+O/IdsBcblvnd8qcO/l1TYoNeAzp3wcISOxlPXQEus0mLcdwazEkWjaBdaJ3TaxmNgCLWwvWzg==",
+      "integrity": "sha1-hNt+nrVTHfGKjF4L+25EnlXmVLI=",
       "dev": true,
       "requires": {
         "samsam": "1.3.0"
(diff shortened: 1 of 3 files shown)


silto:
I've made a fix to the webdav-server code to avoid decoding paths more than once on the server side. The issue was pretty much the same, the path went through the decoder more than once which caused paths with special characters to be impossible to list.
Link to my pending PR:
--> https://github.com/OpenMarshal/npm-WebDAV-Server/pull/33
This fixes the tests with this PR (texted locally). I'll bump the webdav server version once a release is created and this should be all good!

perry-mitchell:
Ah excellent! Thanks for following this through.. 👍

perry-mitchell:
Hi @silto - Any chance you might be able to check that this works with the latest webdav server version?

silto:
@perry-mitchell this works with the updated webdav server package. The tests fail because I didn't update the package-lock.json file (I use old npm & node versions that I definitely need to update), if you can do it the issue would be solved.

perry-mitchell:
Excellent, thanks @silto !

Issue #40, opened 2017-06-22
digest authentication support


tjaartvdwalt:
Hi, looks like only Basic authentication is currently supported? Would it be possible/easy to add Digest authentication?

I am trying to connect to a server that does not appear to support Basic Auth. I get the following error.

```
No 'Authorization: Digest' header found. Either the client didn't send one, or the server is misconfigured
```

perry-mitchell:
Thanks @tjaartvdwalt - I suppose it wouldn't be too difficult to add. I'll try to look into it soon.

tjaartvdwalt:
@perry-mitchell, thanks for the feedback. 
I might look into this more to get my app working correctly. If I do, I will send a pull request.

perry-mitchell:
That would be of course greatly appreciated.. good luck!

perry-mitchell:
Hi @tjaartvdwalt - Did you have any luck here? If not I might look into it this week.

zh99998:
there is a header generator
https://github.com/inorganik/digest-auth-request

perry-mitchell:
@zh99998 Thanks for the link. This library looks great, however it uses its own request mechanism (XMLHttpRequest) under the hood, which is not compatible with out use case. This could be ported to this library, however, by thoroughly running through their [source code](https://github.com/inorganik/digest-auth-request/blob/master/digestAuthRequest.js).

tjaartvdwalt:
I have have done some work on this. The code is probably 80% there, but maybe not production ready.

I will submit a pull request, then we can discuss further.

tjaartvdwalt:
**Update:** my branch fell a bit behind, and I have quite a few merges I need to handle. Will try to get it fixed after work

emilian:
Any updates on digest authentication? Has anybody implemented it?

perry-mitchell:
@emilian I might get a chance to try this today.. installing Ubuntu server to setup digest auth as we speak. Have been meaning to get to it for a while now.

emilian:
Yes, the above tool looks pretty decent to handle the digest authentication.

In the end I went with [this lower level library ](https://github.com/vinger4/node-http-digest-client) to PUT a file on the WebDAV server. It is the only library that worked, since it seems that the request library only works with GET requests out of the box.

It would be awesome if this library could handle digest authentication since it supports a lot of the WebDAV features that will come in handy later on.

perry-mitchell:
[...]

PR #96, opened 2018-05-22
Support for digest authentication

jbruni:
Implements #40

### DIFF
--- a/source/auth.js
+++ b/source/auth.js
@@ -1,13 +1,80 @@
 "use strict";
 
-function generateBasicAuthHeader(username, password) {
+const crypto = require("crypto");
+
+let username, password;
+
+function generateBasicAuthHeader(user, pass) {
+    username = user;
+    password = pass;
     return "Basic " + Buffer.from(username + ":" + password).toString("base64");
 }
+
 function generateTokenAuthHeader(tokenInfo) {
     return `${tokenInfo.token_type} ${tokenInfo.access_token}`;
 }
 
+function generateDigestAuthHeader(url, options, digest) {
+    function md5(data) {
+        return crypto
+            .createHash("md5")
+            .update(data)
+            .digest("hex");
+    }
+
+    function ha1Compute(algorithm, user, realm, pass, nonce, cnonce) {
+        const ha1 = md5(`${user}:${realm}:${pass}`);
+        if (algorithm && algorithm.toLowerCase() === "md5-sess") {
+            return md5(`${ha1}:${nonce}:${cnonce}`);
+        } else {
+            return ha1;
+        }
+    }
+
+    const _url = url.replace("//", "");
+    const uri = _url.indexOf("/") == -1 ? "/" : _url.slice(_url.indexOf("/"));
+
+    const method = options.method ? options.method.toUpperCase() : "GET";
+
+    const qop = /(^|,)\s*auth\s*($|,)/.test(digest.qop) && "auth";
+    const ncString = `00000000${digest.nc}`.slice(-8);
+    const cnonce = digest.cnonce;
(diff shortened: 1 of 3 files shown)


jbruni:
@perry-mitchell 

- It does not use `crypto-js` package. It uses built-in NodeJS `crypto` functions.
- In fact, it does not add any dependency.
- It uses any `fetchMethod` (the default one, `node-fetch`, or any other specified using `setFetchMethod`.
- I have taken the "core idea" of "decorating" `node-fetch` from [`digest-fetch`](https://github.com/devfans/digest-fetch).
- But improved its code borrowing from [`request`](https://github.com/request/request) implementation of digest authentication.

I think the end result is robust. If it can't merge as it is, I hope it serves as a starting point for your own improvements.

Of course, I hope digest authentication support lands as soon as possible, so we can use the official package rather than my fork in our project.

Thanks a lot!

source/auth.js
```diff
@@ -1,13 +1,80 @@
 "use strict";
 
-function generateBasicAuthHeader(username, password) {
+const crypto = require("crypto");
+
+let username, password;
```
perry-mitchell: Not sure I love the idea of these being outside of function scope..

source/auth.js
```diff
@@ -1,13 +1,80 @@
...
+            return ha1;
+        }
+    }
+
+    const _url = url.replace("//", "");
+    const uri = _url.indexOf("/") == -1 ? "/" : _url.slice(_url.indexOf("/"));
+
+    const method = options.method ? options.method.toUpperCase() : "GET";
+
+    const qop = /(^|,)\s*auth\s*($|,)/.test(digest.qop) && "auth";
```
perry-mitchell: Could this be simplified to `\bauth\b`, for instance?

Also I'd prefer it if the value was explicit here.. like `? "auth" : null;` for instance..

source/fetch.js
```diff
@@ -0,0 +1,73 @@
+"use strict";
+
+const { generateDigestAuthHeader } = require("./auth.js");
+const { merge } = require("./merge.js");
+
+const digest = { nc: 0, algorithm: "md5" };
+
+let hasDigestAuth = false;
```
perry-mitchell: Having these as statics doesn't allow for multiple webdav-client instances.. Perhaps we could somehow tie all of this to the webdav instance out of the factory?

perry-mitchell:
Couple of comments @jbruni - Had some ideas about changes that might better protect webdav-client in the cases where it is instantiated multiple times. Great work though, really looking forward to merging this! 😁

jbruni:
@perry-mitchell - 10 months later... I started updating the PR for current `webdav-client` master branch, taking into consideration your feedback. It shall be ready today or tomorrow.

source/auth.js
jbruni: See https://github.com/perry-mitchell/webdav-client/pull/152/commits/8c52e4ffd8d7265abcc4f01c6e956ef861cb80c8#diff-1995eac40392e12643bbf44abad3cf8c

source/auth.js
jbruni: See https://github.com/perry-mitchell/webdav-client/pull/152/commits/8c52e4ffd8d7265abcc4f01c6e956ef861cb80c8#diff-1995eac40392e12643bbf44abad3cf8cR35

I kept the bulk of the regular expression, though, because I'm not its author and I am not sure if that is handling edge cases.

source/fetch.js
jbruni: See https://github.com/perry-mitchell/webdav-client/pull/152/commits/8c52e4ffd8d7265abcc4f01c6e956ef861cb80c8#diff-4970d419691cca822418c65c1fe218ff

jbruni:
Closing in favor of #152

Issue #104, opened 2018-09-13
[1.5.4] path.posix breaks in browserify


perry-mitchell:
> This should be solved another way. PR completely broke the ability to use webdav in the browser. It would be great to have checks for this sort of stuff.

> path-browserify doesn't implement posix yet browserify/path-browserify#11

perry-mitchell:
Considering using the https://www.npmjs.com/package/path-posix library to solve this issue.

perry-mitchell:
The more I read on this the better this idea seems.. Using this library will circumvent the need for the node package. It does require the `util` node package, but from the sounds of things this should build in webpack just fine.

perry-mitchell:
Released in 1.5.5

PR #105, opened 2018-09-13
Use path-posix

perry-mitchell:
Use the [path-posix](https://www.npmjs.com/package/path-posix) library to fix issue #104 where node's path.posix is not implemented in browserify.


perry-mitchell:
Added a bonus feature: Webpack test for confirming that compilation will succeed in bundlers.

Issue #99, opened 2018-08-24
Use Axios


perry-mitchell:
[Axios](https://github.com/axios/axios) is an XMLHttpRequest wrapper for Node and the browser. The benefit here is that it doesn't use fetch and should be much easier to configure cross-platform.

PR #108, opened 2018-09-15
Expose advanced fetch() options

subdavis:
Options provided to public API were not passed through to fetch.

Firefox has some bugs in the fetch() behavior that cause it to contradict [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch) when run inside the context of a browser extension.  For context, this PR will help me solve https://github.com/subdavis/Tusk/issues/210

In order to get desired behavior, I need to be able to pass `mode` and `credentials` arguments on a per-request basis.  This PR re-factors the options argument to a totally pass-through behavior, so whatever a user dumps into the optional `options` map will go to node-fetch or window.fetch


perry-mitchell:
Hi @subdavis - Generally I encourage contributors to make new issues to discuss larger changes like this, rather than having them do potentially fruitless work due to some different direction I'm taking. In issue #99 I mentioned the desire to move to [Axios](https://github.com/axios/axios) for all HTTP requests as indeed fetch is broken all over the place.

I use this library in Node, Chrome, Firefox, Opera, Electron/Nightmare and React-Native.. Fetch is a huge pain as it is never implemented exactly the same way in every case. Browser implementations are mostly ok, but Node (node-fetch as is currently used) vs React-Native is especially painful to manage.

Axios will replace node-fetch, which will unfortunately make this PR obsolete I'm afraid.

perry-mitchell:
By the way, that Nextcloud issue sounds very similar to the problems we're having with Buttercup and Nextcloud/ownCloud due to CORS and other things. Fetch is broken on some cases within the browser as it sends a null Origin field which breaks the CORS implementation on these services.

Axios was one solution to bypass this pain as XHRs give us a little more control.

subdavis:
That's fine, this PR was really an auxiliary goal -- I just wanted to open because I'd rather keep them in-sync with the master library.  These fixes solve a pretty critical issue for me and I need to get them out the door ASAP, so I didn't really have the time to wait for an axios rewrite.

For the record, I don't really consider this a major change.  All I did was re-scope where `merge()` happened.  I suspect your issues with NextCloud/ownCloud are identical to the ones my users are seeing.  Feel free to follow that issue, I should have the fix later today.

perry-mitchell:
@subdavis See [this issue](https://github.com/buttercup/buttercup-browser-extension/issues/140) for our CORS problems.. We've been following alongside the ownCloud guys for some time on this issue. They should've released a fix already, though I'm not sure if or when this'll reach Nextcloud downstream..

perry-mitchell:
I think I'll migrate to Axios quite soon then.. It could end up helping quote a few people to do so.. Was considering making a major revision if this is the case as I don't think an engine change is appropriate for a minor semver change.

PR #111, opened 2018-10-19
Ability to handle write stream errors

ofstudio:
Handle write stream errors `someWriteStream.on('error', ...`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -872,13 +872,13 @@
     },
     "babel-plugin-syntax-async-functions": {
       "version": "6.13.0",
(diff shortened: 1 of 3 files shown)


perry-mitchell:
Thanks @ofstudio! I'll need to get the tests passing before we can merge this. Of course the failure is not due to your changes - It actually looks like the testing dep `nock` has [deprecated node less than 6](https://github.com/nock/nock/releases).

I can update this later, but please feel free to do it yourself in your branch if you want to speed up the process. Nock@9.6.1 should fix it.

PR #114, opened 2018-11-05
V2

### LINKED ISSUE #13: Return values of getRequests should be an Object with a content Field instead of pure content-value
Return values of `getRequests` should be an Object with a content Field instead of pure content-value.
If the `getRequest`-Methodes return an object like 
```
{
     content: <value>
}
```
instead of `<value>`, additional return values may be provided (like response headers).
Such addional values may be `requested` via the options param (see #12 ).

This change would break backwards compatibility.
Alternatively new function-names may be used for this `new return method`, so that compatibility is granted.

### LINKED ISSUE #16: Get headers from all requests
Provide a way to get content **and headers** from every request.

### LINKED ISSUE #99: Use Axios

### LINKED ISSUE #113: webdav throws self.fetch.bind error on iOS
I've created an Ionic 3 App and want to use WebDAV to connect with my server.
When I want to access the module and start the App on an iOS device, I get following error:
`undefined is not an object (self.fetch.bind)`

I access the module following way:
```
import * as WebDAV from 'webdav';

@Component({
    selector: 'page-login',
    templateUrl: 'login.html'
})

export class LoginPage {

    cunstructor() {
        var webdav = WebDAV;  // On this code snippet the error appears
    }
}
```

I've found out, that the error happens in the module node-fetch/browser.js:
```
"use strict";

module.exports = exports = self.fetch;

// Needed for TypeScript and Webpack
exports.default = self.fetch.bind(self);

exports.Headers = self.Headers;
exports.Request = self.Request;
exports.Response = self.Response;
```

I couldn't figure out any solution or workaround and I'm not sure if this only related to node-fetch or also to the webdav module.

perry-mitchell:
Version 2:

 * `axios` instead of `node-fetch` - Fixes #113, closes #99 
 * Detailed responses - closes #16, closes #13


perry-mitchell:
First stage released as 2.0.0-rc1 on npm.

PR #125, opened 2019-01-02
server creation function naming and log formatting

### LINKED ISSUE #124: naming and other minor issues in test/server/index.js
I was just setting up test/server/index.js in another project to have a webdav server for tests and I noticed some minor things that might be helpful, or not :-)

[test/server.index.js](https://github.com/perry-mitchell/webdav-client/blob/master/test/server/index.js#L49) has the following:

```js
createServer.webdavClient = function(authType) {
    return createServer(path.resolve(__dirname, "../testContents"), authType);
};
```
and I think that this might be better as:

```js
createServer.webdavServer = function(authType) {
    return createServer(path.resolve(__dirname, "../testContents"), authType);
};
```
because it is a server being created here right? not a client?

If we changed that would lead to a corresponding change in [test/specs/index.js](https://github.com/perry-mitchell/webdav-client/blob/master/test/specs/index.js#L6)

```js
const createWebDAVServer = baseWebDAVServer.webdavServer;
```

and similarly in [test/web/server.js](https://github.com/perry-mitchell/webdav-client/blob/master/test/web/server.js#L7)

I think that's the only two places it's being referred to ...

tansaku:
fixes #124

### DIFF
--- a/test/server/index.js
+++ b/test/server/index.js
@@ -46,7 +46,7 @@ createServer.test = {
     port: 9988
 };
 
-createServer.webdavClient = function(authType) {
+createServer.webdavServer = function(authType) {
     return createServer(path.resolve(__dirname, "../testContents"), authType);
 };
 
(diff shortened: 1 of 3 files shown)


perry-mitchell:
Thanks @tansaku !

PR #128, opened 2019-01-04
Append slash for PROPFIND requests

### LINKED ISSUE #121: Error: Unexpected close tag
After upgrading to the latest 2.x version, I saw a strange error. After troubleshooting in the existing project, I decided to create a brand new project with only the following code and the same error persisted. (Note that the actual URL and credentials still work fine in 1.x).

```js
const { createClient } = require('webdav');

const client = createClient(
  'http://example.com/webdav',
  {
    username: 'user',
    password: 'pass'
  }
);

// Get directory contents
client.getDirectoryContents('/').then(files => console.log(files));
```

```
(node:55105) UnhandledPromiseRejectionWarning: Error: Unexpected close tag
Line: 25
Column: 7
Char: >
    at error (/Users/claviska/Projects/webdav-test/node_modules/sax/lib/sax.js:651:10)
    at strictFail (/Users/claviska/Projects/webdav-test/node_modules/sax/lib/sax.js:677:7)
    at closeTag (/Users/claviska/Projects/webdav-test/node_modules/sax/lib/sax.js:871:9)
    at SAXParser.write (/Users/claviska/Projects/webdav-test/node_modules/sax/lib/sax.js:1436:13)
    at Parser.exports.Parser.Parser.parseString (/Users/claviska/Projects/webdav-test/node_modules/xml2js/lib/parser.js:322:31)
    at Parser.parseString (/Users/claviska/Projects/webdav-test/node_modules/xml2js/lib/parser.js:5:59)
    at /Users/claviska/Projects/webdav-test/node_modules/webdav/dist/interface/dav.js:30:16
    at new Promise (<anonymous>)
    at parseXML (/Users/claviska/Projects/webdav-test/node_modules/webdav/dist/interface/dav.js:29:12)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:55105) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:55105) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

It seems like an error parsing the WebDAV server's response, but given that 1.x still works fine I don't think it's anything to do with the server. I'm thinking it's probably related to the switch to Axios.

Thanks for your insight!

Node version: 10.2.1

### LINKED ISSUE #127: Force trailing slash on PROPFIND requests
Some servers require a trailing slash on PROPFIND requests - [RFC 4918](https://tools.ietf.org/html/rfc4918#section-5.2):

> There is a standing convention that when a collection is referred to
by its name without a trailing slash, the server MAY handle the
request as if the trailing slash were present. In this case, it
SHOULD return a Content-Location header in the response, pointing to
the URL ending with the "/". For example, if a client invokes a
method on http://example.com/blah (no trailing slash), the server may
respond as if the operation were invoked on http://example.com/blah
(trailing slash), and should return a Content-Location header with
the value http://example.com/blah. Wherever a server produces a URL
referring to a collection, the server SHOULD include the trailing
slash. In general, clients SHOULD use the trailing slash form of
collection names. If clients do not use the trailing slash form the
client needs to be prepared to see a redirect response.
Clients will find the DAV:resourcetype property more reliable than the URL to find
out if a resource is a collection.

Relates to #121

claviska:
This PR fixes #121 and fixes #127. Unfortunately, one test isn't passing and I'm not sure why.

```
  53 passing (678ms)
  1 failing

  1) getDirectoryContents
       returns only expected results when using trailing slash:

      AssertionError: expected '/webdav,/webdav/server' to equal '/webdav/server'
      + expected - actual

      -/webdav,/webdav/server
      +/webdav/server
```

The WebDAV server does appear to provide different results for both URLs, but on the other hand the updated URL is correct. The difference is:

Without slash:
```js
  { filename: '/webdav/server',
    basename: 'server',
    lastmod: 'Fri, 04 Jan 2019 18:22:54 GMT',
    size: 0,
    type: 'directory' }
```

With slash:
```js
{ filename: '/webdav',
    basename: 'webdav',
    lastmod: 'Fri, 04 Jan 2019 18:22:54 GMT',
    size: 0,
    type: 'directory' },
  { filename: '/webdav/server',
    basename: 'server',
    lastmod: 'Fri, 04 Jan 2019 18:22:54 GMT',
    size: 0,
    type: 'directory' }
```

Is it reporting the current directory as a "self" or "dot" directory? I'm not sure why it would vary based on that slash.

### DIFF
--- a/source/interface/directoryContents.js
+++ b/source/interface/directoryContents.js
@@ -4,12 +4,12 @@ const { merge } = require("../merge.js");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
 const { normaliseHREF, normalisePath } = require("../url.js");
 const { getSingleValue, getValueForKey, parseXML, propsToStat } = require("./dav.js");
-const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
+const { encodePath, prepareRequestOptions, request } = require("../request.js");
 
 function getDirectoryContents(remotePath, options) {
     // Join the URL and path for the request
     const requestOptions = {
-        url: appendSlash(joinURL(options.remoteURL, encodePath(appendSlash(remotePath)))),
+        url: joinURL(options.remoteURL, encodePath(remotePath), "/"),
         method: "PROPFIND",
         headers: {
             Accept: "text/plain",
--- a/source/interface/directoryContents.js
+++ b/source/interface/directoryContents.js
@@ -4,14 +4,12 @@ const { merge } = require("../merge.js");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
 const { normaliseHREF, normalisePath } = require("../url.js");
 const { getSingleValue, getValueForKey, parseXML, propsToStat } = require("./dav.js");
-const { encodePath, prepareRequestOptions, request } = require("../request.js");
+const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
 
-function getDirectoryContents(remotePathRaw, options) {
-    // Strip the ending slash
-    const remotePath = remotePathRaw.replace(/\/$/, "");
+function getDirectoryContents(remotePath, options) {
     // Join the URL and path for the request
     const requestOptions = {
-        url: joinURL(options.remoteURL, encodePath(remotePath)),
+        url: appendSlash(joinURL(options.remoteURL, encodePath(appendSlash(remotePath)))),
         method: "PROPFIND",
         headers: {
             Accept: "text/plain",
--- a/source/interface/quota.js
+++ b/source/interface/quota.js
@@ -1,11 +1,11 @@
 const joinURL = require("url-join");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
-const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
+const { encodePath, prepareRequestOptions, request } = require("../request.js");
 const { getSingleValue, getValueForKey, parseXML, translateDiskSpace } = require("./dav.js");
 
 function getQuota(options) {
     const requestOptions = {
-        url: appendSlash(joinURL(options.remoteURL)),
+        url: joinURL(options.remoteURL, "/"),
         method: "PROPFIND",
         headers: {
             Accept: "text/plain",
--- a/source/interface/quota.js
+++ b/source/interface/quota.js
@@ -1,11 +1,11 @@
 const joinURL = require("url-join");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
-const { encodePath, prepareRequestOptions, request } = require("../request.js");
+const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
 const { getSingleValue, getValueForKey, parseXML, translateDiskSpace } = require("./dav.js");
 
 function getQuota(options) {
     const requestOptions = {
-        url: joinURL(options.remoteURL, "/"),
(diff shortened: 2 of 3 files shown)


perry-mitchell:
> Is it reporting the current directory

Yes this has been a pain point historically. We've had to be very careful to ensure that the base directory isn't included in the results. It'll be to do with some matching of the results to what the base directory is expected to be.

source/interface/directoryContents.js
```diff
@@ -4,14 +4,12 @@ const { merge } = require("../merge.js");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
 const { normaliseHREF, normalisePath } = require("../url.js");
 const { getSingleValue, getValueForKey, parseXML, propsToStat } = require("./dav.js");
-const { encodePath, prepareRequestOptions, request } = require("../request.js");
+const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
```
perry-mitchell: No need for `appendSlash` actually.. I discovered that using the `url-join` library is enough:

eg:

```javascript
const joinURL = require("url-join");

joinURL(options.remoteURL, remotePath, "/"); // Creates URL with trailing slash
```

source/interface/directoryContents.js
```diff
@@ -4,14 +4,12 @@ const { merge } = require("../merge.js");
...
 
-function getDirectoryContents(remotePathRaw, options) {
-    // Strip the ending slash
-    const remotePath = remotePathRaw.replace(/\/$/, "");
+function getDirectoryContents(remotePath, options) {
     // Join the URL and path for the request
     const requestOptions = {
-        url: joinURL(options.remoteURL, encodePath(remotePath)),
+        url: appendSlash(joinURL(options.remoteURL, encodePath(appendSlash(remotePath)))),
```
perry-mitchell: This could actually be `joinURL(options.remoteURL, encodePath(remotePath), "/")`

source/interface/quota.js
```diff
@@ -1,11 +1,11 @@
 const joinURL = require("url-join");
 const { handleResponseCode, processResponsePayload } = require("../response.js");
-const { encodePath, prepareRequestOptions, request } = require("../request.js");
+const { appendSlash, encodePath, prepareRequestOptions, request } = require("../request.js");
 const { getSingleValue, getValueForKey, parseXML, translateDiskSpace } = require("./dav.js");
 
 function getQuota(options) {
     const requestOptions = {
-        url: joinURL(options.remoteURL, "/"),
+        url: appendSlash(joinURL(options.remoteURL)),
```
perry-mitchell: You could actually leave it was the original `joinURL(options.remoteURL, "/")` actually.. This is enough to add the slash.

perry-mitchell:
I made some change requests - It seems that the `joinURL` command is enough to ensure the trailing slash.

source/interface/directoryContents.js
claviska: That was my initial approach, but I thought you asked for a helper function perhaps for clarity. (Sorry for the dup function btw)

claviska:
I've removed `appendSlash` as requested. Unfortunately, I don't have time this week to troubleshoot the issue with the current directory being output.

perry-mitchell:
@claviska Thanks! I can have a looksie perhaps today. Cheers :)

Issue #130, opened 2019-01-15
Propfind Depth Header


skoschik:
First, great library.
One suggestion, I think it would be really helpful to support the 'depth' header for PROPFIND requests.   In particular I'm interested in sending Depth:'Infinity' for my use case.

perry-mitchell:
With regards to your PR, #131, one part is definitely to pass a custom depth but we need to handle the following points:

 * We should validate depth to be a number `>= 1`
 * We should modify the parsing of the `getDirectoryContents` method so that it _recursively_ creates the file structure that's returned

It's important that we handle this recursively as imo it makes little sense to have a tree's depth of files contained on 1 "level". We can, when `depth > 1`, add a `children` array to each directory result and recurse into that. In this regard this issue becomes more complex to implement, but I feel it's the right way to go about handling infinite depth.

We could also consider adding an additional property, `depth`, to each result. Every file and folder could have `depth: n` where `n` is the depth it was fetched at (handled by the recursive function), starting at `1`.

skoschik:
For me it's actually optimal as is, being a flat response.
The WebDAV server's XML response is flat, and I think it's easier to deal with the flat results than a nested result.  I'm not sure how much value there is in creating a nested object from the WebDAV server's response.  Recursion everywhere.
I am syncing data between the client and server and on the client I do a recursive directory read BUT that outputs a flat response which can be easily be compared to the webdav's flat response. 

Would you be willing to entertain the idea of an option to have a nested or flat result?

skoschik:
I opened an [issue](https://github.com/OpenMarshal/npm-WebDAV-Server/issues/48) on npm-WebDAV-servers's git because they haven't supported Depth:Infinity yet.  Although, they seem quite interested in doing so.  


We will need them to support this in order to get tests for this feature.
I'm currently using an Apache ModDAV to test with, since it supports depth:infinity

skoschik:
Depth option of`deep: true` sounds great.

Glad you're okay with the flat return object.  

npm-WebDAV-Server released the update to fix the infinity requests today (version 2.4.3).  So we should be able to implements the appropriate tests after making the changes to the server config.

PR #133, opened 2019-01-15
Etags

### LINKED ISSUE #132: getDirectoryContents: etags in resulting entries
It would be really helpful to have the etags in the propfind responses.  It makes for a quick way to see changes across a lot of files/directories in one shot.  Used in conjunction with the depth header = 'infinite' it's a nice way to only pick out changes without blasting the server with individual requests for things that may not have changed.

skoschik:
Closes #132

### DIFF
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -49,7 +49,7 @@ function propsToStat(props, filename) {
         lastmod: lastMod,
         size: parseInt(rawSize, 10),
         type: type,
-        etag: etag.replace(/"/g, "")
+        etag: (etag) ? etag.replace(/"/g, "") : null
     };
     if (type === "file") {
         stat.mime = mimeType ? mimeType.split(";")[0] : "";
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -42,12 +42,14 @@ function propsToStat(props, filename) {
     const resourceType = getSingleValue(getValueForKey("resourcetype", props));
     const mimeType = getSingleValue(getValueForKey("getcontenttype", props));
     const type = getValueForKey("collection", resourceType) ? "directory" : "file";
+    const etag = getSingleValue(getValueForKey("getetag", props));
     const stat = {
         filename: filename,
         basename: path.basename(filename),
         lastmod: lastMod,
         size: parseInt(rawSize, 10),
-        type: type
+        type: type,
+        etag: etag.replace(/"/g, "")
     };
     if (type === "file") {
         stat.mime = mimeType ? mimeType.split(";")[0] : "";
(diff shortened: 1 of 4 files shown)


source/interface/dav.js
```diff
@@ -42,12 +42,14 @@ function propsToStat(props, filename) {
...
     const type = getValueForKey("collection", resourceType) ? "directory" : "file";
+    const etag = getSingleValue(getValueForKey("getetag", props));
     const stat = {
         filename: filename,
         basename: path.basename(filename),
         lastmod: lastMod,
         size: parseInt(rawSize, 10),
-        type: type
+        type: type,
+        etag: etag.replace(/"/g, "")
```
perry-mitchell: Maybe `etag && .replace(/"/g, "") || null` ?

source/interface/directoryContents.js
```diff
@@ -13,7 +13,7 @@ function getDirectoryContents(remotePath, options) {
         method: "PROPFIND",
         headers: {
             Accept: "text/plain",
-            Depth: 1
+            Depth: options.depth || 1
```
perry-mitchell: This can be removed as it's part of another issue - #130

test/web/specs/getDirectoryContents.spec.js
```diff
@@ -96,4 +96,10 @@ describe("getDirectoryContents", function() {
             expect(contents[0].basename).to.equal("file2.txt");
         });
     });
+
+    it("returns etags from propfind", function() {
+        return this.client.getDirectoryContents("/", { depth: 1 }).then(function(contents) {
+            expect(contents[0]).to.have.property("etag");
```
perry-mitchell: You can specify the etag as a second parameter to ensure the value is correct.

test/web/specs/getDirectoryContents.spec.js
```diff
@@ -96,4 +96,10 @@ describe("getDirectoryContents", function() {
             expect(contents[0].basename).to.equal("file2.txt");
         });
     });
+
+    it("returns etags from propfind", function() {
+        return this.client.getDirectoryContents("/", { depth: 1 }).then(function(contents) {
```
perry-mitchell: Let's remove `depth` - That's part of #130

perry-mitchell:
Hi @skoschik - Just waiting on your response to my review.. If we get those items sorted I can merge. Thanks!

source/interface/dav.js
skoschik: would this work?  not sure if i followed the example you gave exactly.
```
        etag: (etag) ? etag.replace(/"/g, "") : null
```

source/interface/directoryContents.js
skoschik: Got it

source/interface/dav.js
perry-mitchell: That's fine too of course

PR #135, opened 2019-01-19
Add props to detailed responses (#134)

### LINKED ISSUE #134: Allow access to all entry properties
Hi,

is it possible to access WebDAV entry properties that are not directly exposed by the api?
I'm connecting to a server with a compilance class greater than 1 and 2 which returns some more properties that are not part of the WebDAV standard.

Example response:
```XML
    <D:propstat>
      <D:prop>
        <D:isvirtual>true</D:isvirtual>         <-- this one
        <D:resourcetype>
          <D:collection />
        </D:resourcetype>
        <D:getlastmodified>Fri, 18 Jan 2019 15:51:16 GMT</D:getlastmodified>
        <D:creationdate>2019-01-18T15:51:16.106Z</D:creationdate>
        <D:getcontentlength>0</D:getcontentlength>
        <D:lockdiscovery />
        <D:supportedlock />
      </D:prop>
      <D:status>HTTP/1.1 200 OK</D:status>
    </D:propstat>
```

Do you know a way to do this?
If not, I'd be happy to submit a pull request.

**Edit:** Doesn't seem to be possible: https://github.com/perry-mitchell/webdav-client/blob/9944d22400955702da4a8c54b385250d0bd46f1c/source/interface/dav.js#L37
Maybe this could be implemented by returning the raw `props` when `{ details: true }` is set like it's done with the headers?

**Edit 2:** When looking at NextCloud WebDAV responses you can see that they're using custom propierties, too. So this is probably not only for my a relevant feature.

Kind regards,
Marcus

MarcusWichelmann:
When the option `details` is enabled, the stats of each WebDAV file and collection should contain the original props to allow the reading of addtional item information.

This fixes #134.

**Breaking Changes:** None

**New API:**
When the options contain `{ details: true }` the result data of `getDirectoryContents()` and `stat()` contains a new `props` object next to other stat-fields like `basename` and `filename`.

Example response:
```JSON
[
   {
      "filename":"/sub1",
      "basename":"sub1",
      "lastmod":"Sat, 19 Jan 2019 14:06:55 GMT",
      "size":0,
      "type":"directory",
      "props":{
         "D:getlastmodified":[
            "Sat, 19 Jan 2019 14:06:55 GMT"
         ],
         "D:lockdiscovery":[...],
         "D:supportedlock":[...],
         "D:creationdate":[
            "2019-01-19T14:06:55-01:00"
         ],
         "D:resourcetype":[...],
         "D:displayname":[
            "sub1"
         ],
         "D:getetag":[
            "\"ad26f60f92b318bfe4e833069bed23d6\""
         ]
      }
   },
   ...
]
```

**TODO:**
- [ ] Maybe a new unit test should be added to check if new custom properties are received. But I didn't find an easy way to do this using the node `webdav-server` library.

### DIFF
--- a/test/specs/getDirectoryContents.spec.js
+++ b/test/specs/getDirectoryContents.spec.js
@@ -35,6 +35,9 @@ describe("getDirectoryContents", function() {
             expect(details)
                 .to.have.property("headers")
                 .that.is.an("object");
+            expect(details.data[0])
+                .to.have.property("props")
+                .that.is.an("object");
         });
     });
 
--- a/source/interface/directoryContents.js
+++ b/source/interface/directoryContents.js
@@ -26,11 +26,11 @@ function getDirectoryContents(remotePath, options) {
             return res.data;
         })
         .then(parseXML)
-        .then(result => getDirectoryFiles(result, options.remotePath, remotePath))
+        .then(result => getDirectoryFiles(result, options.remotePath, remotePath, options.details))
         .then(files => processResponsePayload(response, files, options.details));
 }
 
-function getDirectoryFiles(result, serverBasePath, requestPath) {
+function getDirectoryFiles(result, serverBasePath, requestPath, isDetailed = false) {
     const remoteTargetPath = pathPosix.join(serverBasePath, requestPath, "/");
     const serverBase = pathPosix.join(serverBasePath, "/");
     // Extract the response items (directory contents)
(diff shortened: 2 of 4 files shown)


test/specs/getDirectoryContents.spec.js
```diff
@@ -35,6 +35,9 @@ describe("getDirectoryContents", function() {
             expect(details)
                 .to.have.property("headers")
                 .that.is.an("object");
+            expect(details.data[0])
```
MarcusWichelmann: @perry-mitchell This makes on of the tests fail. I'm not familiar with this unit test framwork. What's the right way to do this?

perry-mitchell: The only test I can see failing is a nodejs 6 test, but it's failing due to some pathing issues, which I believe were already fixed. Very strange. I'll try to investigate further shortly.

perry-mitchell: I’ll most likely merge this tonight and will resolve any issues I find then.. thanks!

PR #152, opened 2019-03-31
Support for digest authentication

### LINKED ISSUE #40: digest authentication support

jbruni:
Closes #40

### DIFF
--- a/source/auth.js
+++ b/source/auth.js
@@ -1,3 +1,4 @@
+const crypto = require("crypto");
 const { toBase64 } = require("./encode.js");
 
 function generateBasicAuthHeader(username, password) {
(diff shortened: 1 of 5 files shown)


jbruni:
Hi, @perry-mitchell 

In this new PR, I addressed the issues raised at #96, as well as updated the whole solution to fit in the changes done in `webdav-client` since that PR has been created. I have also created a new branch for this PR because updating the other branch would have side-effects for us using it as dependency.

These issues were addressed:

- No "statics outside of function scope" anymore.
- Digest stuff is tied at webdav instance in the factory, at creation time.
- So, multiple webdav-client instances is not an issue anymore.

:warning: The main difference, though, is that now the library consumer must explicitly opt-in the digest authentication usage. I have not updated documentation, though. To opt-in, it is only a matter of adding `digest: true` to the basic auth client options:

```js
const { createClient } = require("webdav");

const client = createClient(
    "https://webdav.example.com/marie123",
    {
        username: "marie",
        password: "myS3curePa$$w0rd",
        digest: true
    }
);
```

Without `digest: true`, there are no changes to the current client behavior.

Please feel free to make any adjustments you want to.

Thanks!

perry-mitchell:
@jbruni This is amazing! Thanks so much for taking the time to see this through, I appreciate it. I'll be sure to give this a once over asap. Fantastic :)

source/auth.js
```diff
@@ -9,7 +10,67 @@ function generateTokenAuthHeader(tokenInfo) {
...
 
+function generateDigestAuthHeader(options, digest) {
+    function md5(data) {
+        return crypto
+            .createHash("md5")
+            .update(data)
+            .digest("hex");
+    }
+
+    function ha1Compute(algorithm, user, realm, pass, nonce, cnonce) {
```
perry-mitchell: Could we take these functions out of `generateDigestAuthHeader` and put them on the top level? It'd be nice to add tests for them at some stage :)

perry-mitchell: `md5` could be taken out too, even though it's quite basic.

perry-mitchell:
Also @jbruni - I wonder if there's some relatively simple way I can test this? I know of no digest-auth based WebDAV services I could try this on.

jbruni:
Hi @perry-mitchell - I can take care of the code change request.

Meanwhile, let me suggest a WebDAV service for you to test. By the way, the same service we are using for real projects.

You can create a BigCommerce free trial account, which comes with a free store: https://www.bigcommerce.com/essentials/pricing/

Every BigCommerce store comes with a WebDAV server access to manage assets: https://support.bigcommerce.com/s/article/File-Access-WebDAV#login-info

These assets may be publicly accessible from HTTP.  For example: a file uploaded as `/content/theme/homepage/highlight-pass-atlantis.jpg` may become available at https://resort-for-a-day-sandbox-1.mybigcommerce.com/content/theme/homepage/highlight-pass-atlantis.jpg (this is valid for `/content` path).

Their WebDAV server software is `sabre/dav` - http://sabre.io/dav/ - and they are probably using one of the two built-in Digest auth classes - http://sabre.io/dav/authentication/

source/auth.js
jbruni: I have moved them to a separate `crypto.js` file, in the same fashion you have a separate `encode.js` file.

(FYI, I amended previous commit only to fix its author/committer.)

perry-mitchell:
Thanks @jbruni so much for this! I haven't tested it yet but will merge it to another temp branch while I do. Great stuff :)

PR #154, opened 2019-04-09
sometimes boolean true is returned for mimetype

wodka:
only add the stat.mime entry if mimetype is a string

### DIFF
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -60,7 +60,7 @@ function propsToStat(props, filename, isDetailed = false) {
         etag: etag ? etag.replace(/"/g, "") : null
     };
     if (type === "file") {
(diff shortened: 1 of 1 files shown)


source/interface/dav.js
```diff
@@ -59,7 +59,7 @@ function propsToStat(props, filename, isDetailed = false) {
         type: type,
         etag: etag ? etag.replace(/"/g, "") : null
     };
-    if (type === "file") {
+    if (type === "file" && typeof mimeType === 'string') {
```
perry-mitchell: This `typeof` would be better added to the next line so that if the mime type is _not_ a string, it returns the empty value from the ternary if.

Oh and please use double quotes :)

wodka: just updated it accordingly

PR #165, opened 2019-07-05
Support properties on propfind

DeepDiver1975:
This adds new options for the method getDirectoryContents to add PROPFIND properties

### DIFF
--- a/source/factory.js
+++ b/source/factory.js
@@ -42,7 +42,10 @@ const stats = require("./interface/stat.js");
 /**
  * @typedef {OptionsForAdvancedResponses} GetDirectoryContentsOptions
  * @property {Boolean=} deep - Return deep (infinite) items (default: false)
+ * @property {Boolean=} details - Response objects hold all properties (default: false)
  * @property {String=} glob - Glob pattern for matching certain files
+ * @property {Array=} properties - Array of properties which shall be use in the PROPFIND request body
+ * @property {Array=} namespaces - Array of XML namespaces which are used in the PROPFIND request body
  */
 
 /**
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -90,10 +90,49 @@ function translateDiskSpace(value) {
     }
 }
 
+function parseClarkNotation(propertyName) {
+    const result = propertyName.match(/^{([^}]+)}(.*)$/);
+    if (!result) {
+        return;
+    }
+
+    return {
+        name: result[2],
+        namespace: result[1]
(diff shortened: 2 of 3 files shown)


DeepDiver1975:
looks like axios does not transmit a body for propfind .... still evaluating ...
https://github.com/axios/axios/issues/2270

perry-mitchell:
Hi @DeepDiver1975! Thanks for looking in to this. Might I ask why you'd ever need this? I haven't yet come across a server that'd need data sent on a PROPFIND request, besides the usual headers. If this functionality alters the response body in the wrong way, it'd be quite easy to generate unsupported result structures (as in unsupported by the parsing logic in this client).

DeepDiver1975:
Webdav, caldav and carddav have a big number of properties per spec which can be queried for. In addition any property can be set via propset and using propfind these props can also be queried for.

perry-mitchell:
@DeepDiver1975 Makes sense. Could you attempt to add a test for this new functionality, however? I don't feel comfortable adding new functionality which isn't demonstrated or tested, sorry.

We use `webdav-server` as a dependency which should allow you to write tests against it for most webdav-supported functionality.

perry-mitchell:
@DeepDiver1975 Would you have any time to add the tests soon? I'm preparing a new major version of this library centered around a new requests engine I've written: [cowl](https://github.com/perry-mitchell/cowl). There's no restrictions on PROPFIND requests as you mentioned with axios, so I'm not sure what you might recommend in this case. 

The new major will be in RC mode for a while in a separate branch - I'll link it here shortly.

DeepDiver1975:
> @DeepDiver1975 Would you have any time to add the tests soon? I'm preparing a new major version of this library centered around a new requests engine I've written: [cowl](https://github.com/perry-mitchell/cowl). There's no restrictions on PROPFIND requests as you mentioned with axios, so I'm not sure what you might recommend in this case.

I did not put any prio to this as long as axios is not capable of submitting PROPFIND.

DeepDiver1975:
@perry-mitchell any news on the branch using cowl? I'd like to give that a try ... THX

perry-mitchell:
@DeepDiver1975 Work on cowl stalled a bit, though I do intend on continuing. Issue is that cowl needs **streams** to be a good fit for this library, and streams don't work with XHRs. So I need to update cowl to use:
 * http/https clients in node
 * fetch in browser
 * fetch in react-native ?

So that'll require a bit more fiddling I think.

PR #168, opened 2019-07-09
Support additional body in requests

yoh1496:
This PR make it possible to send requests with the `body` parameter to WebDav server.

The `prepareRequestOptions` function ignores the `body` parameter in methodOptions.
So, this PR modifies codes to avoid ignoring the parameter and makes options compatible with `axios` options by renaming `body` to `data`.

Please review this.


perry-mitchell:
This might conflict with #165 - which came first. So if that gets merged we might need to revisit this MR.

perry-mitchell:
Webdav might have some serious components rewritten soon, such as the requests library portion or potentially even the whole library. I'll close this as it's not exactly how I'd want to integrate more custom body requests. Thanks.

Issue #147, opened 2019-02-15
How to know when writeStream is completed?


sean256:
I need to know when a `createWriteStream` has fully completed.

I tried `on` events for `end`, `finish` and `unpipe` and all of them get called immediately and before the write is complete.

```
const readStream = fs.createReadStream('./test.zip');
const writeStream = client.createWriteStream('/some/remote/place/test.zip');
writeStream.on('end', () => console.log('end'))
		.on('unpipe', () => console.log('unpipe'))
		.on('finish', () => console.log('finish'));
readStream.pipe(writeStream);
```

sean256:
I found a way to get a proper event when it's done but requires a code change in webdav-client

in `createStream.js` -> `createWriteStream` I added a line to emit to `writeStream` when request does it's thing.

```
prepareRequestOptions(requestOptions, options);
    request(requestOptions)
        .then(responseHandlers.handleResponseCode)
        .then(() => {
            writeStream.emit("complete");
        })
        .catch(err => {
            writeStream.emit("error", err);
        });
    return writeStream;
```

perry-mitchell:
I'm not sure what `"complete"` would do.. Aren't the stream events only `close`, `open` and `ready` ([`WriteSteam`](https://nodejs.org/api/fs.html#fs_class_fs_writestream))?

sean256:
You're right. Here is what I need to know and it looks like one of your tests have the same need.

I need to know when the "WriteStream" is totally and completely done doing its thing, including passing off to `request` and `request` completing. My current use case is in AWS Lambda and if you return in the `handle` function AWS may freeze your lambda until is awaken again even if there is still stuff to do in the event loop!

Besides my use in AWS Lambda I think knowing when it's completed is a good thing to have anyway. I just don't know what the best option is here with the way it's set up.

```
// here's a snippet of what I'm doing using the proposed `completed`. This is wrapped up in a promise.
const readStream = s3.getObject({ Bucket, Key }).createReadStream().on('error', reject);
const writeStream = webdavClient.createWriteStream(`${remotePath}/${filename}`);
readStream.pipe(writeStream);
writeStream.on('complete', resolve); // resolve promise
```

perry-mitchell:
I've flagged this as a bug, which I'll hopefully get around to shortly. I'm also happy for someone else to take a look at it!

bingliu221:
Same issue here.
[...]

PR #176, opened 2019-10-06
Discussed fix for #147

Glebka:
So I created a PR that implements what was discussed in #147:
I added an ability to pass Readable stream as data argument + created a test for it.

@perry-mitchell, please review.


perry-mitchell:
Hi @Glebka - Thanks for this! Seems the tests are failing but are mostly unrelated.. That being said, I don't have time right now to find the issue.. Would there be any chance you might take a took?

Seems to be xvfb failing on web, which I can try restarting now, and seems that some dependency updated on the node side (Hapi rubbish most likely) to add `...` spreading, which doesn't work on Node 6.

I'll be supporting node 6 until a new version is released - soon.

perry-mitchell:
Ok, upon closer look I'm not sure that this will work for downstream users in the browser. I think it's best that we don't use native node modules, especially streams and other file/IO items, as they don't transpile or transpile poorly.

I'll leave this open for dicussion, but I'm not sure how we can move forward with this in its current state.

perry-mitchell:
@Glebka I've updated the PR to include checks for the new web version, hope you don't mind. Waiting to see if the tests pass now.

PR #180, opened 2019-10-29
Forward cancelToken to request options

### LINKED ISSUE #179: Cancelable requests
Passing a proper cancelToken to a `getDirectoryContents` options does not work.
If I just replace the `getDirectoryContents` function by a simple axios.get function, it works.

So I'm assuming there is something here that block the passing of the data to the final axios library?

``` js
import axios from 'axios'

/**
 * Creates a cancelable axios 'request object'.
 *
 * @param {function} request the axios promise request
 * @returns {Object}
 */
const CancelableRequest = function(request) {
	/**
	 * Generate an axios cancel token
	 */
	const CancelToken = axios.CancelToken
	const source = CancelToken.source()

	/**
	 * Execute the request
	 *
	 * @param {string} url the url to send the request to
	 * @param {Object} [options] optional config for the request
	 */
	const fetch = async function(url, options) {
		return request(
			url,
			Object.assign({ cancelToken: source.token }, { options })
		)
	}
	return {
		request: fetch,
		cancel: source.cancel
	}
}

export default CancelableRequest
```

``` js
const { request, cancel } = cancelableRequest(webdav.getDirectoryContents)
// enable slow network to have the time to cancel the request or add some delay to your webdav backend
request()
setTimeout(() => {
	cancel('Stop!')
}, 100)
```

skjnldsv:
Fix #179

Though I cannot catch the error properly.
I can when using `axios.get`, but I cannot with the webdav lib.

Axios is outputing this in the console:
Capture d’écran_2019-10-29_15-38-48
But webdav is throwing
Capture d’écran_2019-10-29_15-39-05

### DIFF
--- a/source/request.js
+++ b/source/request.js
@@ -51,6 +51,9 @@ function prepareRequestOptions(requestOptions, methodOptions) {
     if (typeof methodOptions.withCredentials === "boolean") {
         requestOptions.withCredentials = methodOptions.withCredentials;
     }
+    if (methodOptions.cancelToken && typeof methodOptions.cancelToken === "object") {
+        requestOptions.cancelToken = merge(requestOptions.cancelToken || {}, methodOptions.cancelToken);
+    }
     if (methodOptions.maxContentLength) {
         requestOptions.maxContentLength = methodOptions.maxContentLength;
     }


perry-mitchell:
Please see comments in #179 - I think more planning is required before moving forward with this. I don't want to pass in options to axios directly if I can avoid it, as I want the API to be very much separate. I plan to change the request engine in the near future.

skjnldsv:
@perry-mitchell then may I suggest to just extend all the options and let axios do whatever it wants with it?
I would actually think that is the proper way to do it since you're plainly using axios, so if you want to be future proof, then let's just do this.

But I did not want to break the pattern you were already using :)

perry-mitchell:
That’s the thing, I **dont’t** want to use axios.. and I wouldn’t think it a good idea anyway to just pass through options - making the API dependent on the engine used. It should be entirely independent.

skjnldsv:
Let's continue in here: https://github.com/perry-mitchell/webdav-client/issues/179 :)

perry-mitchell:
As discussed in #179, we agreed that a cancel token is a fine way to go, but it has to be created by WebDAV (to future-proof the API).

@skjnldsv let me know if you'd want to add such a function, otherwise we should probably close this and come back to it later.

skjnldsv:
@perry-mitchell thanks :)
Well, I don't really have the time for ti unfortunately, so let's close in favor of 179

PR #195, opened 2020-02-02
Check for NULL props

### LINKED ISSUE #194: Many props undefined when getting directory contents from NextCloud webdav
Webdav: 3.0.0
NextCloud: 16.0.1

As shown in the code below, I have commented out most of this function to get directory listing in NextCloud. I'm not really sure if this is a NextCloud issue or not.

``` bash
 15 function propsToStat(props, filename, isDetailed = false) {
 16   const path = require("path-posix"); // Last modified time, raw size, item type and mime
 17 
 18 
 19   //const lastMod = props.getlastmodified,
 20   //      _props$getcontentleng = props.getcontentlength,
 21   //      rawSize = _props$getcontentleng === void 0 ? "0" : _props$getcontentleng,
 22   //      resourceType = props.resourcetype,
 23   //      mimeType = props.getcontenttype,
 24   //      etag = props.getetag;
 25   //const type = resourceType && typeof resourceType === "object" && typeof resourceType.collection !== "undefined" ? "directory" : "file";
 26   const stat = {
 27     filename: filename,
 28     basename: path.basename(filename),
 29     //lastmod: lastMod,
 30     //size: parseInt(rawSize, 10),
 31     //type: type,
 32     //etag: typeof etag === "string" ? etag.replace(/"/g, "") : null
 33   };
 34 
 35   //if (type === "file") {
 36   //  stat.mime = mimeType && typeof mimeType === "string" ? mimeType.split(";")[0] : "";
 37   //}
 38 
 39   if (isDetailed) {
 40     stat.props = props;
 41   }
 42 
 43   return stat;
 44 }
```

Dogcatfee:
fixes #194

### DIFF
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -19,7 +19,14 @@ function propsToStat(props, filename, isDetailed = false) {
         resourcetype: resourceType,
         getcontenttype: mimeType,
         getetag: etag
-    } = props;
+    } = props ? props : 
+    {
+      getlastmodified: null, 
+      getcontentlength: null, 
+      resourceType: null, 
+      getcontenttype: null, 
+      getetag: null
+    };
     const type =
         resourceType && typeof resourceType === "object" && typeof resourceType.collection !== "undefined"
             ? "directory"
@@ -35,7 +42,7 @@ function propsToStat(props, filename, isDetailed = false) {
     if (type === "file") {
         stat.mime = mimeType && typeof mimeType === "string" ? mimeType.split(";")[0] : "";
     }
-    if (isDetailed) {
+    if (props && isDetailed) {
         stat.props = props;
     }
     return stat;


skjnldsv:
@perry-mitchell :)

source/interface/dav.js
```diff
@@ -19,7 +19,14 @@ function propsToStat(props, filename, isDetailed = false) {
         resourcetype: resourceType,
         getcontenttype: mimeType,
         getetag: etag
-    } = props;
+    } = props ? props :
```
perry-mitchell: Shouldn't props always be an object here? I don't think it should be optional.

source/interface/dav.js
```diff
@@ -19,7 +19,14 @@ function propsToStat(props, filename, isDetailed = false) {
         resourcetype: resourceType,
         getcontenttype: mimeType,
         getetag: etag
```
perry-mitchell: Instead of setting them to a new object below, why not do something like this:

```javscript
const {
	getlastmodified: lastMod = null,
    getcontentlength: rawSize = "0",
    resourcetype: resourceType = null
   // ... etc
} = props;
```

That way you don't need to copy the entire object below.

Dogcatfee: Good idea, I had no idea you could do something like that. But unfortunately that doesn't appear to work.

What I changed it to:
``` javascript
 16     const {
 17         getlastmodified: lastMod = null,
 18         getcontentlength: rawSize = "0",
 19         resourcetype: resourceType = null,
 20         getcontenttype: mimeType = null,
 21         getetag: etag = null
 22     } = props;
```

What gets generated:
``` javascript
 19   const _props$getlastmodifie = props.getlastmodified,
 20         lastMod = _props$getlastmodifie === void 0 ? null : _props$getlastmodifie,
 21         _props$getcontentleng = props.getcontentlength,
 22         rawSize = _props$getcontentleng === void 0 ? "0" : _props$getcontentleng,
 23         _props$resourcetype = props.resourcetype,
 24         resourceType = _props$resourcetype === void 0 ? null : _props$resourcetype,
 25         _props$getcontenttype = props.getcontenttype,
 26         mimeType = _props$getcontenttype === void 0 ? null : _props$getcontenttype,
 27         _props$getetag = props.getetag,
 28         etag = _props$getetag === void 0 ? null : _props$getetag;
 29   const type = resourceType && typeof resourceType === "object" && typeof resourceType.collection !== "undefined" ? "directory" : "file";
```

source/interface/dav.js
Dogcatfee: Props is somehow null in this case

source/interface/dav.js
```diff
@@ -35,7 +42,7 @@ function propsToStat(props, filename, isDetailed = false) {
     if (type === "file") {
         stat.mime = mimeType && typeof mimeType === "string" ? mimeType.split(";")[0] : "";
     }
-    if (isDetailed) {
+    if (props && isDetailed) {
```
Dogcatfee: This change is unnecessary

perry-mitchell:
As we’ve been speaking about this in #196, I don’t think we need this MR. The new changes should ensure that props is set, and they now have defaults. Thanks :)

PR #200, opened 2020-03-23
provide callback for writeStream

h-bar:
Currently there's no way to see if the request sent by writeStream (for uploading a file) is finished or not. The writeStream.on("finish") event is fired before the request is complement since the request itself is synchronize. This change adds a callback option when creating writeStream so that the callback is executed when the request sent by the writeStream is finished sending data

### DIFF
--- a/source/factory.js
+++ b/source/factory.js
@@ -174,19 +174,20 @@ function createClient(remoteURL, opts = {}) {
          * Create a writeable stream to a remote file
          * @param {String} remoteFilename The file to write to
          * @param {PutOptions=} options Options for the request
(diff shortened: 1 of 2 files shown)


perry-mitchell:
@h-bar Thanks! I made some changes to improve the callback handling, as the response handler still needs to be fired.

h-bar:
@perry-mitchell  Thanks so much for your improvement. This is my first merge request on Github so I'm really glad you liked it!

perry-mitchell:
Good work @h-bar - appreciate the MR. Please drop by again if you have more suggestions :) - This is now live in 3.3.0.

PR #201, opened 2020-03-31
Fix merging objects with prototypes

aleksandryackovlev:
The `merge` function in `/source/merge.js` uses `Object.assign` method.
The issue is that this method copies only an object's own properties to the resulting object.

It means we've got an error from https://github.com/nodejs/node/blob/master/lib/_http_client.js#L135 when we pass `httpsAgent` or `httpAgent` as an option to any method of the client.

For example, the following code will throw th an error.
```
await client.putFileContents(filePath, fileBuffer, { httpsAgent: new https.Agent({ rejectUnauthorized: true })});
```

To fix this issue we have to not only copy its own properties of the original object but to copy its prototype too.


source/merge.js
```diff
@@ -1,10 +1,14 @@
+function clone(obj) {
+    return Object.setPrototypeOf(Object.assign({}, obj), Object.getPrototypeOf(obj));
```
perry-mitchell: `setPrototypeOf` is supposedly a very heavy operation:

image

Can we do something like:

```javascript
const OBJ_PROTO = Object.getPrototypeOf({});

function clone(obj) {
	const objProto = Object.getPrototypeOf(obj);
    return objProto === OBJ_PROTO
		? Object.assign({}, obj)
		: Object.setPrototypeOf(Object.assign({}, obj), objProto);
}
```

perry-mitchell:
Hmm.. very interesting. I made a suggestion, to improve the cloning process, but I simply wonder if a helper like the following would be better:

```javascript
function isPureObject(input) {
    if (input === null || typeof input === "undefined") return false;
    return Object.getPrototypeOf(input).isPrototypeOf(Object);
}
```

We could use this to _not_ merge objects that are non-object types (classes, like HttpsAgent).

We'd need to also test for `Array.isArray()` as `[]` doesn't satisfy `isPureObject` of course.

aleksandryackovlev:
Maybe it's going to be better to combine both approaches.
It seems like in lodash.merge they use both of them.
https://github.com/lodash/lodash/blob/master/.internal/baseMergeDeep.js

perry-mitchell:
Thanks @aleksandryackovlev! This is great. Will merge shortly.

PR #213, opened 2020-08-29
Fix getDirectoryContents for directories with & in name

christiaan:
Webdav servers turn & into &amp;amp; in their responses however the client
does not decode this entity.


perry-mitchell (approved):
👍

PR #214, opened 2020-08-29
Add missing documentation for maxContentLength option

christiaan:
This option was added in #146 but not documented.


perry-mitchell:
@christiaan Thank you!

Issue #144, opened 2019-01-30
Support uploading bigger files with createWriteStream


Paxa:
axios uses follow-redirect, which limit body size to 10 MB (for some unknown reasons)
So webdav-client uploading is limited to 10 MB also, and there is no way to specify custom limit:

This example doesn't work because only `httpAgent`, `httpsAgent`, `headers` and `withCredentials` allowed
```js
client.createWriteStream(davPath, {maxContentLength: 1024 ** 3}); // 1 GB
// Error: Request body larger than maxBodyLength limit
```

I think webdav-client should allow any size for uploading files or make it configurable, may be something like global axios settings

perry-mitchell:
Yeah this seems to be an axios flaw like you said. Would be good to add it as an option in this library, with it being auto-set at say 1GB by default. As you showed setting `maxContentLength` in axios options will do the trick.

sean256:
@perry-mitchell I submitted a PR for exactly this

perry-mitchell:
This should be fixed by #146

perry-mitchell:
Released in 2.6.0

PR #215, opened 2020-09-03
Fix unnecessary excessive memory usage when uploading big files

christiaan:
When uploading a file the complete contents of the file is loaded in memory
just in case the request ends up with a redirect response and the request
has to be resent.

see https://github.com/axios/axios/issues/2783

### DIFF
--- a/source/interface/createStream.js
+++ b/source/interface/createStream.js
@@ -29,7 +29,8 @@ function createWriteStream(filePath, options, callback = NOOP) {
         url: joinURL(options.remoteURL, encodePath(filePath)),
         method: "PUT",
         headers,
-        data: writeStream
+        data: writeStream,
+        maxRedirects: 0
     };
     prepareRequestOptions(requestOptions, options);
     request(requestOptions)


christiaan:
As far as I could see and know Webdav does not user redirect headers, especially not when uploading a file.

This change breaks BC, is this acceptable? Or should this be turned into an option?

I could not imagine a situation where you want to support uploading all your files twice without being notified about it.

christiaan:
BTW this also fixes https://github.com/perry-mitchell/webdav-client/issues/144 by removing the need to set `maxContentLength` because the `follow-redirects` module is completely disabled by `maxRedirects: 0`.

christiaan:
@perry-mitchell is there anything I can help with to move this forward?

perry-mitchell:
@christiaan I'm sorry, I actually completely forgot about this, my bad! There's nothing holding this back, and I can move it forward. I'll release it under a feature as I'm not sure that any potential issues will be immediately apparent. Thanks a lot for this one!

perry-mitchell (approved):

PR #219, opened 2020-10-12
Fix empty multistatus content

### LINKED ISSUE #218: Error: Invalid response: No root multistatus found
```xml
<?xml version="1.0"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns" />

```

skjnldsv:
Fix #218

cc @perry-mitchell @juliushaertl @christophwurst

### DIFF
--- a/source/interface/dav.js
+++ b/source/interface/dav.js
@@ -21,6 +21,13 @@ function normaliseResponse(response) {
 
 function normaliseResult(result) {
     const { multistatus } = result;
+    if (multistatus === '') {
+        return {
+            multistatus: {
+                response: []
+            }
+        }
+    }
     if (!multistatus) {
         throw new Error("Invalid response: No root multistatus found");
     }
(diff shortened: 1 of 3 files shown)


ChristophWurst (approved):

source/interface/dav.js
```diff
@@ -21,6 +21,13 @@ function normaliseResponse(response) {
 
 function normaliseResult(result) {
     const { multistatus } = result;
+    if (multistatus === '') {
```
perry-mitchell, line 24: Would this be better as a `!multistatus` check? This would catch occurences where `multistatus` might not be returned at all..

perry-mitchell (changes requested):

source/interface/dav.js
skjnldsv: well, it is bellow :)

skjnldsv: But bellow it throw, which is not good for empty multistatus. :thinking:

perry-mitchell (approved):

skjnldsv:
Thanks a lot @perry-mitchell !
Love your lib :)

emansom:
Great job, my savior! Is this merge included in latest Nextcloud?

PR #220, opened 2020-10-12
Allow setting the option `maxBodyLength` on `putFileContents()`

bennigraf:
Hi there,

I recently ran into a problem when uploading big files to a nextcloud instance. Axios would complain

```
Error [ERR_FR_MAX_BODY_LENGTH_EXCEEDED]: Request body larger than maxBodyLength limit
```

This PR allows using the option `maxBodyLength` directly on the `putFileContents` method of this webdav client. One would use it for example like this:

```
await client.putFileContents(
  "/remote/file", 
  fs.readFileSync('./a-big-file'), 
  { 
    'maxBodyLength': Infinity,
  }
);
```

To achieve this, the option is simply passed through to axios which in turn will pass it on to `follow-redirects`, which is it's default backend in nodejs environments. It did in fact solve the issue in my project.

Until earlier this year, axios would have used the value of `maxContentLength` to set the value of `maxBodyLength`, but at some point someone found out that those two are in fact not the same and changed that behaviour, which I assume broke functionality of `webdav-client` as well. In my understanding, `maxContentLength` is now supposed to be used only for setting the limit of the *respnse* size, while `maxBodyLength` defines the limit of the *request* body size. See this issue for reference: https://github.com/axios/axios/issues/2696

Let me know what you think!

Best,
Benjamin.

### DIFF
--- a/source/factory.js
+++ b/source/factory.js
@@ -25,7 +25,8 @@ const NOOP = () => {};
  * Options for creating a resource
  * @typedef {UserOptions} PutOptions
  * @property {Boolean=} overwrite - Whether or not to overwrite existing files (default: true)
- * @property {Number=} maxContentLength - The maximum amount of bytes to upload (default: 10 MiB)
+ * @property {Number=} maxContentLength - The maximum amount of bytes to download (default: 10 MiB)
+ * @property {Number=} maxBodyLength - The maximum amount of bytes to upload (default: 10 MiB) (only in nodejs-environments)
  */
 
 /**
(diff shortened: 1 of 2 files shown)


perry-mitchell:
@bennigraf Have you tried with 3.5.0? Shouldn't [this PR](https://github.com/perry-mitchell/webdav-client/pull/215#issuecomment-686452181) have fixed it? :)

perry-mitchell (approved):

bennigraf:
Awesome, thanks for the quick approval and for releasing this directly!

And thanks alot for creating and maintaining this nice library. 🥇

PR #226, opened 2020-11-07
Move to webpack 5

skjnldsv:
To be adjusted.
I'm not sure how this is supposed to behave for the node build. There is maybe no need for the polyfill then.
Only for the umd browser version this should be needed :thinking: 

cc @perry-mitchell

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "nyc": "^15.1.0",
     "prettier": "^1.19.1",
     "prettier-check": "^2.0.0",
+    "process": "^0.11.10",
     "rimraf": "^3.0.2",
     "sinon": "^9.2.0",
     "wait-on": "^5.2.0",
--- a/package.json
+++ b/package.json
@@ -53,14 +53,19 @@
   "dependencies": {
     "axios": "^0.20.0",
     "base-64": "^0.1.0",
+    "buffer": "^6.0.1",
+    "crypto-browserify": "^3.12.0",
     "fast-xml-parser": "^3.17.4",
     "he": "^1.2.0",
     "hot-patcher": "^0.5.0",
     "minimatch": "^3.0.4",
     "nested-property": "^4.0.0",
+    "path-browserify": "^1.0.1",
     "path-posix": "^1.0.0",
+    "stream-browserify": "^3.0.0",
     "url-join": "^4.0.1",
-    "url-parse": "^1.4.7"
+    "url-parse": "^1.4.7",
+    "util": "^0.12.3"
   },
   "devDependencies": {
     "@babel/cli": "^7.11.6",
@@ -92,7 +97,7 @@
     "sinon": "^9.2.0",
     "wait-on": "^5.2.0",
     "webdav-server": "^2.6.2",
-    "webpack": "^4.44.2",
-    "webpack-cli": "^3.3.12"
+    "webpack": "^5.4.0",
+    "webpack-cli": "^4.2.0"
   }
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1279,6 +1279,44 @@
         "defer-to-connect": "^1.0.1"
       }
     },
+    "@types/eslint": {
+      "version": "7.2.4",
+      "resolved": "https://registry.npmjs.org/@types/eslint/-/eslint-7.2.4.tgz",
+      "integrity": "sha512-YCY4kzHMsHoyKspQH+nwSe+70Kep7Vjt2X+dZe5Vs2vkRudqtoFoUIv1RlJmZB8Hbp7McneupoZij4PadxsK5Q==",
+      "dev": true,
+      "requires": {
+        "@types/estree": "*",
+        "@types/json-schema": "*"
+      }
+    },
+    "@types/eslint-scope": {
+      "version": "3.7.0",
+      "resolved": "https://registry.npmjs.org/@types/eslint-scope/-/eslint-scope-3.7.0.tgz",
+      "integrity": "sha512-O/ql2+rrCUe2W2rs7wMR+GqPRcgB6UiqN5RhrR5xruFlY7l9YLMn0ZkDzjoHLeiFkR8MCQZVudUuuvQ2BLC9Qw==",
+      "dev": true,
+      "requires": {
+        "@types/eslint": "*",
+        "@types/estree": "*"
+      }
+    },
+    "@types/estree": {
+      "version": "0.0.45",
+      "resolved": "https://registry.npmjs.org/@types/estree/-/estree-0.0.45.tgz",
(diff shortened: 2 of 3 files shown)


perry-mitchell:
I'm not going to merge this in its current form for several reasons:

 * Tests are failing
 * Unneeded dependencies added (buffer etc.)

That, and I'm in no hurry to upgrade to webpack 5. Webpack 4 works perfectly fine. I don't see any benefits of upgrading to v5, so please let me know if I'm missing something.

skjnldsv:
> * Unneeded dependencies added (buffer etc.)

It's not, webpack used to ship them and provided them by default, with the version 5 they stopped. And ask you to add them yourself :wink: 



> * Tests are failing

Well, for this, definitely!
I couldn't open a draft here, but I'm definitely not done :)

skjnldsv:
> That, and I'm in no hurry to upgrade to webpack 5. Webpack 4 works perfectly fine. I don't see any benefits of upgrading to v5, so please let me know if I'm missing something.

You could say that to a lot of dependencies.
And then you're stuck because everyone upgraded, or an old lib fails and your upgrade path gets insanely more complicated.

In any case:
https://webpack.js.org/blog/2020-10-10-webpack-5-release/#general-direction

> # This release focus on the following:
> 
> - Improve build performance with Persistent Caching.
> - Improve Long Term Caching with better algorithms and defaults.
> - Improve bundle size with better Tree Shaking and Code Generation.
> - Improve compatibility with the web platform.
> - Clean up internal structures that were left in a weird state while implementing features in v4 without introducing any breaking changes.
> - Prepare for future features by introducing breaking changes now, allowing us to stay on v5 for as long as possible.

skjnldsv:
> Sorry to come off as short - I appreciate the PR. But everything needs to be rock solid as this package has such a large number of dependents. Unnecessary modifications should be avoided where possible.

Also, no worries at all! I'm 100% on your side regarding this package. Let's be up-to-date and make sure we keep on improving on a steady basis. Rushing is not an option, and staying behind should not be either. :)

perry-mitchell:
> It's not, webpack used to ship them and provided them by default, with the version 5 they stopped. And ask you to add them yourself

I see, I hadn't read this. Well, that's fair enough, but `buffer` is built in.. Not sure why we need to install a package for it.. Unless that's required to bundle it. If that's to happen, they should at least be in the devDependencies, as NodeJS doesn't need them.

I also see that you added `process` to the deps.. this again is another Node built-in as you know so I'm not sure why we need to install a dep of the same name - If Webpack 5 requires this it sounds like a broken methodology to me.

> And then you're stuck because everyone upgraded, or an old lib fails and your upgrade path gets insanely more complicated. 

You're making the assumption that an older Webpack version will eventually break dependents.. but it won't. Webpack is a build-time dependency, and its version has no effect on the consumers of this library. I could still be using webpack 2 with no negative affects on packages requiring the node or web versions :)

My point is that there absolutely isn't a reason to rush to WP5 in this case.. Webpack 4 has been configured to output an optimised bundle for this library and upgrading to WP5 will not drastically improve this process.

> Rushing is not an option, and staying behind should not be either. :)

Agreed, though we're _far_ from being behind. I'd say even another 12 months on v4 would be fine, considering the size of the webpack ecosystem. I wouldn't plan on waiting 12 months, it's just an example.

-------------

I digress - For me to be comfortable merging this I'd ideally want the following:

 * No extra dependencies, say beyond `crypto-browserify`, added to npm
 * Extra size tests added to master (I'll try to take care of this asap)
   * No size increase, not even a little bit - from using WP5
 * Unchanged compatibility (browsers) - Not sure how to check this one, but it's required

Happy to discuss alternatives to what I'm suggesting.

package.json
```diff
@@ -53,14 +53,19 @@
...
     "axios": "^0.20.0",
     "base-64": "^0.1.0",
+    "buffer": "^6.0.1",
+    "crypto-browserify": "^3.12.0",
     "fast-xml-parser": "^3.17.4",
     "he": "^1.2.0",
     "hot-patcher": "^0.5.0",
     "minimatch": "^3.0.4",
     "nested-property": "^4.0.0",
+    "path-browserify": "^1.0.1",
```
perry-mitchell, line 63: This should be removed - we have `path-posix` which should be enough.

package.json
```diff
@@ -53,14 +53,19 @@
   "dependencies": {
     "axios": "^0.20.0",
     "base-64": "^0.1.0",
+    "buffer": "^6.0.1",
```
perry-mitchell, line 56: This should be removed - see discussion

package.json
```diff
@@ -53,14 +53,19 @@
...
     "hot-patcher": "^0.5.0",
     "minimatch": "^3.0.4",
     "nested-property": "^4.0.0",
+    "path-browserify": "^1.0.1",
     "path-posix": "^1.0.0",
+    "stream-browserify": "^3.0.0",
     "url-join": "^4.0.1",
-    "url-parse": "^1.4.7"
+    "url-parse": "^1.4.7",
+    "util": "^0.12.3"
```
perry-mitchell, line 68: This should be removed - see discussion

package.json
```diff
@@ -88,11 +93,12 @@
     "nyc": "^15.1.0",
     "prettier": "^1.19.1",
     "prettier-check": "^2.0.0",
+    "process": "^0.11.10",
```
perry-mitchell, line 96: This should be removed - see discussion

perry-mitchell (changes requested):

skjnldsv:
> Agreed, though we're _far_ from being behind. I'd say even another 12 months on v4 would be fine, considering the size of the webpack ecosystem. I wouldn't plan on waiting 12 months, it's just an example.

Definitely! It's just a matter of time spans. No need to rush for webpack 5. I don't even remember why I jumped into this pr to be honest, I've been context switching a lot over the last weeks ^^
This pr isn't on my priorities right now, so having it opened and waiting a bit is perfectly fine to me ;)

skjnldsv:
> * No extra dependencies, say beyond `crypto-browserify`, added to npm

The fact that webpack asked me to add them means propably that one of the deps you're using is also using them.
If that is the case you won't have a choice to also have them if they don't bundle the deps :/

perry-mitchell:
Well, I definitely can’t comment conclusively here as I haven’t yet used webpack 5.  I can’t imagine though that it’d require installing dependencies like buffer and process from **npm**.. they’re built ins and it seems weird that you’d have to provide some unassociated third party library.

Anywho perhaps there’s a way to either:

- Remove the dependencies that use buffer or process (use webpack to set them to empty)
- Or poly fill them with custom libraries like path-posix, that I’m already using. My guess is here that it’s possible, looking at the code you’ve added..

Like you said no need to rush.. perhaps I’ll check out your PR soon and will take a look myself..

perry-mitchell:
I'll close this as I'll handle the migration to v5 myself I think - want to pay special attention to the bundling in doing so. Hope you understand :)

skjnldsv:
Of course! :)
Good luck! :v:

PR #228, opened 2020-11-16
Update README.md

bennigraf:
Hi,

I'd like to propose to document how to specify the request range when creating a read stream. I had to dig into the source code to find out how to do this; I thought it might help others to have this available in the readme.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,17 +144,17 @@ Create a read stream targeted at a remote file:
 ```javascript
 client
     .createReadStream("/video.mp4")
-    .pipe(fs.createWriteStream("~/video.np4"));
+    .pipe(fs.createWriteStream("~/video.mp4"));
 ```
 
-If you want to request only partial data, you can specify the range in the options argument of the method:
+If you want to stream only part of the file, you can specify the `range` in the options argument:
 
 ```javascript
 client
     .createReadStream(
         "/video.mp4", 
         { range: { start: start, end: end } }
-    ).pipe(fs.createWriteStream("~/video.np4"));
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,16 @@ client
...
 ```
 
+If you want to request only partial data, you can specify the range in the options argument of the method:
+
+```javascript
+client
+    .createReadStream(
+        "/video.mp4", 
+        { range: { start: start, end: end } }
+    ).pipe(fs.createWriteStream("~/video.np4"));
```
perry-mitchell, line 157: Should be `.mp4`

README.md
```diff
@@ -147,6 +147,16 @@ client
     .pipe(fs.createWriteStream("~/video.np4"));
 ```
 
+If you want to request only partial data, you can specify the range in the options argument of the method:
```
perry-mitchell, line 150: Could you change this to: "If you want to stream only part of the file, you can specify the `range` in the options argument:"

perry-mitchell (changes requested):

perry-mitchell:
@bennigraf Great, thanks! Just a minor change :)

README.md
bennigraf: Sure!

README.md
bennigraf: Yep, sounds better

bennigraf:
Thanks for your review, I updated the PR according to your requests. Thanks for _your_ work on `webdav-client` :-)

perry-mitchell (approved):

PR #231, opened 2020-12-08
Update custom.js

anatolyukropov:
This is a custom method, so i think he must give me an oppotunity to work with files. 
For example this method set props: 
const contents = await this.webDavClient.customRequest(path, {
        method: 'PROPPATCH',
        headers: {
          Accept: '*/*',
        },
        data: propertyupdate.toXML(),
        responseType: 'application/xml; charset="utf-8"',
      });
But couse of '/' i can set props only to dirs. 
Can you remove this join '/', or make some props like, file with file = true remove '/'

### DIFF
--- a/source/interface/custom.js
+++ b/source/interface/custom.js
@@ -3,7 +3,7 @@ const { encodePath, joinURL, prepareRequestOptions, request } = require("../requ
 
 function customRequest(remotePath, requestOptions, options) {
     if (!requestOptions.url) {
-        requestOptions.url = joinURL(options.remoteURL, encodePath(remotePath), "/");
+        requestOptions.url = joinURL(options.remoteURL, encodePath(remotePath));
     }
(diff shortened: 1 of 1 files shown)


perry-mitchell:
I agree that the forced addition of `/` looks weird, especially for a custom request. `remotePath` could easily include the ending slash if required. Not sure I want to rush into this change as it could have downstream consequences. I'll see if I can't find the prior contributors for this feature to ask them about their usage.

perry-mitchell:
@ubarsaiyan FYI, if you have a moment..

perry-mitchell:
Upon closer inspection it seems like this _was_ added over a year ago.. probably not fresh in anyone's minds right now. Let me have a think on it and will get back to it asap.

perry-mitchell (approved):
Approving this as it blocks accessing files, as mentioned.

PR #250, opened 2021-03-16
Assign onUploadProgress for axios.

tmpfs:
Fixes a bug with upload progress handlers not firing when using `putFileContents`.


source/request.ts
```diff
@@ -48,6 +48,9 @@ export function prepareRequestOptions(
     if (context.maxBodyLength) {
         finalOptions.maxBodyLength = context.maxBodyLength;
     }
+    if (userOptions.hasOwnProperty('onUploadProgress')) {
+        finalOptions.onUploadProgress = userOptions['onUploadProgress'];
```
perry-mitchell, line 52: Could you please change this to use double quotes?

source/types.ts
```diff
@@ -125,6 +125,7 @@ export interface OAuthToken {
 export interface PutFileContentsOptions extends WebDAVMethodOptions {
     contentLength?: boolean | number;
     overwrite?: boolean;
+    onUploadProgress?: UploadProgressCallback;
```
perry-mitchell, line 128: Could you please document this property in the readme?

package-lock.json
```diff
@@ -1,13131 +1,8 @@
 {
   "name": "webdav",
   "version": "4.2.1",
-  "lockfileVersion": 2,
```
perry-mitchell, line 4: Can you please revert the entire package-lock change? It's far too large a change for no dependencies having been updated.

perry-mitchell (changes requested):

source/request.ts
perry-mitchell: Or better yet, add this as an optional property to user options and just use `if (userOptions.onUploadProgress)` etc. :)

package-lock.json
perry-mitchell: If you update from master this is already handled - the lock file was an old version and now it should more closely match what you submitted.

PR #251, opened 2021-03-16
Remove manual assignment of Content-Length header

tmpfs:
Manually attempting to assign `Content-Length` is a flawed strategy as the browser should do this automatically based on the content supplied to `XMLHttpRequest.send()` (for security reasons).

All tests continue to pass so I assume it has not affected the nodejs support.


perry-mitchell:
Some services require this, and Node does not set it automatically. The browser does, of course, in many cases. It's also required when streaming contents - as the length cannot be known beforehand.

perry-mitchell:
Merging this would break a lot of downstream use-cases, so I'll close this.. sorry.

perry-mitchell:
If you feel that there's a bug introduced by how this is implemented, please open an issue so we can discuss it before implementing anything. The code you removed handles the web/node differences.

PR #258, opened 2021-05-05
Consider `status` in `propstat` when doing `stat`

jstastny:
## Why
When `PROPFIND` is sent to Nginx webdav server on a non-existent file, it responds with `207 Multi-Status`, but contains the `404` in the `status` element of `propstat`. This is an example response from Nginx:
```xml
<?xml version="1.0" encoding="utf-8" ?>
<D:multistatus xmlns:D="DAV:">
    <D:response>
        <D:href>/p-nice/does-not-exist</D:href>
        <D:propstat>
            <D:prop></D:prop>
            <D:status>HTTP/1.1 404 Not Found</D:status>
        </D:propstat>
    </D:response>
</D:multistatus>
```

The [RFC 4918, Section 9.1.2.](https://tools.ietf.org/html/rfc4918#section-9.1.2) names the `404` status as valid value for the `status` element in `propstat`, but it does not have an example with this combination (`207` response status and `404` in the `status` element).

## Before
Before this PR, the client would treat the `PROPFIND` responses for non-existent resources from NGinx as if the resources existed. Not considering the `status` element of `propstat` in the response body.

## After
Throw error when doing `stat` on non-existent resource even if the HTTP status of the response is `207`, but the `status` in `propstat` element is `>= 400`.

## Notes
We have been using this code at Deepnote as a package patch for couple months already without any issues, but this was still using the non-TypeScript version of this library. I have waited with creation of the PR before the TypeScript version was out (thank you very much for the TS version, very appreciated).

The same approach could in theory be applied on all `Multi-Status` responses to `PROPFIND` -- I can namely think of the `getDirectoryContents` function.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,7 @@ When making a pull request for an existing issue or feature that is clearly bene
 When making a pull request for an item not listed as an issue or that is controversial (also subject to opinion), be aware that the request may be denied and the PR closed. It is always best to create an issue for a discussion before making a PR, if you are unsure of the feature's worth.
 
 ## Documentation
-PRs to add/update documentation are welcome, but please, ensure that you edit in **Markdown** with as little HTML as possible (none is preferrable). Do not use auto-formatters to generate the documentation files.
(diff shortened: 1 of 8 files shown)


perry-mitchell (approved):

perry-mitchell:
Thanks @jstastny - This is fantastic!

PR #259, opened 2021-05-06
Ensure proper collection path when creating directory

jstastny:
According to the specification (http://www.webdav.org/specs/rfc4918.html#rfc.section.5.2) the collection URI contains the trailing slash. While the specification allows that some implementations may treat the URI without the trailing slash as if it had it, some implementations (nginx namely) is strict about this and only supports the URIs with the trailing slash.

This ensures the URI has the trailing slash before sending it via `MKCOL`.


perry-mitchell (approved):

perry-mitchell:
Great! :)

PR #262, opened 2021-05-13
Add response to the WebDAVClientError type

dpakach:
Add response property to WebDAVClientError type.

the error handleing is no longer handled by axios (https://github.com/perry-mitchell/webdav-client/blob/master/source/request.ts#L52) so the error object doesn't has access to the HTTP response. this PR fixes that.

### DIFF
--- a/source/response.ts
+++ b/source/response.ts
@@ -1,10 +1,9 @@
 import minimatch from "minimatch";
+import { AxiosResponse } from "axios";
 import { FileStat, Response, ResponseDataDetailed, WebDAVClientContext, WebDAVClientError } from "./types";
-import {AxiosResponse} from "axios";
 
 export function handleResponseCode(context: WebDAVClientContext, response: Response): Response {
     const status = response.status;
-
     if (status === 401 && context.digest) return response;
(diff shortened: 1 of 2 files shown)


source/response.ts
```diff
@@ -7,6 +8,7 @@ export function handleResponseCode(context: WebDAVClientContext, response: Respo
     if (status >= 400) {
         const err: WebDAVClientError = new Error(`Invalid response: ${status} ${response.statusText}`) as WebDAVClientError;
         err.status = status;
+        err.response = response as AxiosResponse;
```
perry-mitchell, line 11: I'd prefer not to expose Axios-specific properties outside of this library, as I can't guarantee that I'll always use Axios as the request client. Would you consider changing this to `err.response = response` (only) and using the `Response` type in the `WebDAVClientError` interface?

perry-mitchell (changes requested):

dpakach:
@perry-mitchell I've fixed it. Please have a look.

perry-mitchell (approved):

PR #263, opened 2021-05-16
Fix uploadProgress

### LINKED ISSUE #249: putFileContents does not pass onUploadProgress to axios
Just spent a lot of time debugging why my progress handler was not firing and finally figured out that `putFileContents` is not passing `onUploadProgress` to axios.

I am not sure what the correct fix for this is but happy to submit a *quick* fix (directly assign to `requestOptions`) it you would take the PR.

My guess is the bug is in `prepareRequestOptions()`.

n-peugnet:
I just rebased commits from #250 and dropped the lockfile changes.

I tested it locally and it worked nicely.

Fixes #249 
Supersedes #250

### DIFF
--- a/source/operations/putFileContents.ts
+++ b/source/operations/putFileContents.ts
@@ -35,7 +35,7 @@ export async function putFileContents(
     if (!overwrite) {
         headers["If-None-Match"] = "*";
     }
-    const requestOptions = prepareRequestOptions({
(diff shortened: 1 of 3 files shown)


perry-mitchell (approved):

perry-mitchell:
Thanks @n-peugnet !

PR #291, opened 2022-02-17
Correctly calculate byte size of the utf8 strings

jstastny:
The original implementation started causing problems when the webdav client was used in combination with axios 0.22 and later, as in 0.22 axios [started considering the `Content-Length` passed in the request](https://github.com/axios/axios/pull/2880/files#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fR89-R91) where it previously ignored it and calculated the length on its own.

### DIFF
--- a/source/tools/size.ts
+++ b/source/tools/size.ts
@@ -9,7 +9,7 @@ export function calculateDataLength(data: string | BufferLike): number {
     } else if (isBuffer(data)) {
         return (<Buffer>data).length;
     } else if (typeof data === "string") {
-        return (<string>data).length;
+        return Buffer.byteLength(<string>data);
     }
     throw new Layerr(
         {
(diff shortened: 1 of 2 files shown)


source/tools/size.ts
```diff
@@ -9,7 +9,7 @@ export function calculateDataLength(data: string | BufferLike): number {
     } else if (isBuffer(data)) {
         return (<Buffer>data).length;
     } else if (typeof data === "string") {
-        return (<string>data).length;
+        return Buffer.byteLength(<string>data);
```
perry-mitchell, line 12: Unfortunately this won't work, as `Buffer` is not available in the web build.

perry-mitchell (changes requested):

perry-mitchell:
Seeing as I didn't get any response, I'll merge and modify this fix to use a library which covers byte-size calculations.

jstastny:
> 

Hi.

Sorry for not responding. I stopped hoping that this could go live as is, so I've given up.

Thanks for taking the extra step to make it compatible with browser.

PR #307, opened 2022-06-19
AxiosAdapter option added

MrRefactoring:
Use case:

I'm creating a WebDAV GUI client based on electron with an enabled `nodeIntegration` flag. It means I can use both requests ("browser" and "NodeJS" based).

I wasted a lot of time understanding why `webdav-client` works incorrectly with streams (`fs.createReadStream()`) and when I tried to send file using axios with `agent: require('axios/lib/adapters/http')` option everything started works fine.

So this PR adding `adapter` option support out of the box

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.log
 .DS_Store
 .history
+.idea
 .vscode
 .nyc_output
 test.js
--- a/README.md
+++ b/README.md
@@ -141,18 +141,18 @@ The `createClient` method takes a WebDAV service URL, and a configuration option
 The available configuration options are as follows:
 
 | Option             | Default | Description                                                                                                                               |
-|--------------------|--------|-------------------------------------------------------------------------------------------------------------------------------------------|
-| `adapter`          | _None_ | `adapter` allows custom handling of requests. See [AxiosAdapter](https://github.com/axios/axios/blob/v1.x/lib/adapters/README.md)         |
-| `authType`         | `null` | The authentication type to use. If not provided, defaults to trying to detect based upon whether `username` and `password` were provided. |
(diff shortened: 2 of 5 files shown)


perry-mitchell:
Hi @MrRefactoring - Why not just override the agent as specified in the [docs](https://github.com/perry-mitchell/webdav-client#client-configuration)?

I don't expose Axios because it's just a utility.. I don't want the client to be available/known externally as it should be inconsequential to the API. I might very well switch Axios for something else soon enough, like fetch. So in that way I can't accept this PR as it is.

I'd really like to understand what the actual issue is (streams work fine for me on node), and perhaps address it in another matter. Were you trying to use streams on the browser side? Or are there some other options you want to override on the agent?

hocgin:
Hi @perry-mitchell , Can I now replace axios with fetch in a specific way?
My environment is Chrome Manifest V3, it does not support XMLHttpRequest

perry-mitchell:
Hi @hocgin - Apologies for my silence here, I forgot about this PR.

For you or others that see this, you can bypass needing axios by updating the `HotPatcher` [here](https://github.com/perry-mitchell/webdav-client/blob/master/source/index.ts#L2). You can do _something_ like this:

```typescript
import { RequestOptions, getPatcher } from "webdav";
import fetch from "cross-fetch";

getPatcher().patch(
    "request",
    (opts: RequestOptions) => fetch(opts.url, {
        method: opts.method,
        headers: opts.headers,
        body: opts.data
    });
);
```

That should stop `webdav` from using axios and it should instead use cross-fetch. Note that this doesn't cater for _all_ properties, so you should check out [this interface](https://github.com/perry-mitchell/webdav-client/blob/master/source/types.ts#L157).

PR #309, opened 2022-07-10
feat: add ts overload to getFileContents

magicdawn:
so that we get explicit return type when different options provided

```ts
getFileContents({} as any, "/").then(res => res.byteLength); // res: BufferLike

getFileContents({} as any, "/", { format: "text" }).then(res => res.length); // res: string

getFileContents({} as any, "/", { details: true }).then(res => res.data.byteLength); // res: Detail<BufferLike>

getFileContents({} as any, "/", { details: true, format: "text" }).then(res => res.data.length); // res: Detail<string>
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,7 @@
 end_of_line = lf
 insert_final_newline = true
 
-[*.{js,json}]
+[*.{js,json,ts}]
 indent_style = space
 indent_size = 4
 
--- a/source/operations/getFileContents.ts
+++ b/source/operations/getFileContents.ts
@@ -15,6 +15,30 @@ import {
 
 const TRANSFORM_RETAIN_FORMAT = (v: any) => v;
 
+type GetFileContentsBasicOptions = Omit<GetFileContentsOptions, "format" | "details">;
+
+export async function getFileContents(
+    context: WebDAVClientContext,
+    filePath: string,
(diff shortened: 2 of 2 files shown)


perry-mitchell (approved):
Awesome, thank you!

perry-mitchell:
@magicdawn I had to revert this as types broke after merging:

```
source/factory.ts:94:48 - error TS2769: No overload matches this call.
  The last overload gave the following error.
    Argument of type 'GetFileContentsOptions' is not assignable to parameter of type 'GetFileContentsBasicOptions & { format: "text"; details: true; }'.
      Type 'GetFileContentsOptions' is not assignable to type '{ format: "text"; details: true; }'.
        Types of property 'format' are incompatible.
          Type '"text" | "binary"' is not assignable to type '"text"'.
            Type '"binary"' is not assignable to type '"text"'.

94             getFileContents(context, filename, options),
                                                  ~~~~~~~

  source/operations/getFileContents.ts:36:23
    36 export async function getFileContents(
                             ~~~~~~~~~~~~~~~
    The last overload is declared here.
  source/operations/getFileContents.ts:42:23
    42 export async function getFileContents(
                             ~~~~~~~~~~~~~~~
    The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.


Found 1 error in source/factory.ts:94
```

PR #315, opened 2022-08-19
Added type declaration file for hot-patcher package

santiagocano08:
It is added a type declaration file in the used functions of the "hot-pacher" package so that errors are not generated when the ts files are compiled with strict mode active


perry-mitchell (approved):
Thanks! This'll do until hot patcher is converted. Cheers

PR #318, opened 2022-09-06
CORS header fix coming up in owncloud 10.11.0

jnweiger:
It is fixed in ownCloud 10.11.0 via https://github.com/owncloud/core/pull/40204
Public release is expected next week.


perry-mitchell:
Thanks for this @jnweiger, but I'm still skeptical that this will fix the issue. The PR seems to only fix the parsing of allowed domains, but for CORS to truly work properly there'd have to be an option to allow either all origins, or to somehow support non-standard origins like Chrome/Firefox extensions, for example.

As I've stopped using ownCloud, would it be possible at all to ask you to perhaps run a quick test when it's released? If you're using OC?

perry-mitchell:
Closed due to not receiving a response. I don't have the time to confirm if it's fixed or not, and even if it is I'd want the text to state that it's fixed from version X onwards I think. Too much to confirm here to just merge this I'm afraid.

PR #320, opened 2022-09-16
Handle URL encoded paths returned in directory listings

lyonzy:
I'm using the Synology (Apache based) WebDAV server and it returns characters like `[` and `]` URL encoded:

Path: `/home/[Reference]/dn-test/test/`
WebDAV response: `<D:href>/home/%5bReference%5d/dn-test/test/</D:href>`

This causes the `pathPosix.relative` call to think it's in a different dir, and the paths all come out prepended with `/../../`. Decoding the href value before using it means it comes out correctly (same as others without those characters). Not sure whether this technically belongs in `normaliseHREF` though.

In the meantime I have a polyfill which tackles it at the `normaliseHREF` side:
```
import * as wurl from "webdav/dist/node/tools/url";
var old = wurl.normaliseHREF;
(wurl as any).normaliseHREF = (href: string) => {
    return decodeURIComponent(old(href));
};
```

### DIFF
--- a/source/operations/directoryContents.ts
+++ b/source/operations/directoryContents.ts
@@ -55,8 +55,6 @@ function getDirectoryFiles(
         responseItems
             // Map all items to a consistent output structure (results)
             .map(item => {
-                // HREF is the file path (in full)
-                const href = normaliseHREF(item.href);
                 // HREF is the file path (in full).
                 // decodeURIComponent caters for paths that are returned encoded
                 const href = decodeURIComponent(normaliseHREF(item.href));
--- a/source/operations/directoryContents.ts
+++ b/source/operations/directoryContents.ts
@@ -57,6 +57,10 @@ function getDirectoryFiles(
             .map(item => {
                 // HREF is the file path (in full)
                 const href = normaliseHREF(item.href);
+                // HREF is the file path (in full).
+                // decodeURIComponent caters for paths that are returned encoded
+                const href = decodeURIComponent(normaliseHREF(item.href));
+
                 // Each item should contain a stat object
                 const {
                     propstat: { prop: props }


perry-mitchell:
@lyonzy Thanks for the PR, however you can see that the tests are now failing as your change breaks existing functionality. I'd guess that whatever Synology is doing, they're bending the standard. I'm no expert here however.

Unfortunately I can't merge this as is. Would you be able to share a full response from the PROPFIND request for directory contents? You should of course obscure sensitive values before doing so. That might help us figure out how to get around this particular case.

lyonzy:
Sure all good. Here's an abbreviated PROPFIND response including a directory `/home/notes/[Unfiled]`:

```
<?xml version="1.0" encoding="utf-8"?>
<D:multistatus xmlns:D="DAV:">
    ...
    <D:response xmlns:lp1="DAV:"
        xmlns:lp2="http://apache.org/dav/props/">
        <D:href>/home/notes/%5bUnfiled%5d/</D:href>
        <D:propstat>
            <D:prop>
                <lp1:resourcetype>
                    <D:collection/>
                </lp1:resourcetype>
                <lp1:creationdate>2022-10-03T01:24:58Z</lp1:creationdate>
                <lp1:getlastmodified>Mon, 03 Oct 2022 01:24:58 GMT</lp1:getlastmodified>
                <lp1:getetag>"1000-5ea173641fbd1"</lp1:getetag>
                <D:supportedlock>
                    <D:lockentry>
                        <D:lockscope>
                            <D:exclusive/>
                        </D:lockscope>
                        <D:locktype>
                            <D:write/>
                        </D:locktype>
                    </D:lockentry>
                    <D:lockentry>
                        <D:lockscope>
                            <D:shared/>
                        </D:lockscope>
                        <D:locktype>
                            <D:write/>
                        </D:locktype>
                    </D:lockentry>
                </D:supportedlock>
                <D:lockdiscovery/>
                <D:getcontenttype>httpd/unix-directory</D:getcontenttype>
            </D:prop>
            <D:status>HTTP/1.1 200 OK</D:status>
        </D:propstat>
    </D:response>
    ...
</D:multistatus>

```

I'm no expert on Apache or WebDAV but I know it's an Apache based implementation with mod_dav - here's some relevant parts of httpd.conf:
```
LoadModule dav_module modules/mod_dav.so
LoadModule dav_fs_module modules/mod_dav_fs.so
LoadModule dav_lock_module modules/mod_dav_lock.so

...

<Location />
  Dav on
  DavDepthInfinity on
  AuthType Basic
  AuthName "SYNO_WebDAV Storage"
  AuthBasicProvider socache PAM
  AuthnCacheProvideFor PAM
  AuthPAMService webdav
  Require valid-user
  <IfModule authn_anon_module>
    Anonymous anonymous ftp
    Anonymous_NoUserID off
    Anonymous_MustGiveEmail off
    Anonymous_VerifyEmail off
    Anonymous_LogEmail on
  </IfModule>
</Location>

```

Hope that helps - I'll poke around the standard and see if I can find a definitive answer for whether to expect encoding.

lyonzy:
This comment:
https://datatracker.ietf.org/doc/html/rfc4918#section-8.3.1
Seems to suggest that given the `href` field is a URI reference it needs to follow URI standards (link to the relevant RFC in the text) which includes encoding - and one of their examples uses `%20` for a space.

susnux:
I agree this sometimes causes issues, as the href is defined as an `URI`: https://datatracker.ietf.org/doc/html/rfc4918#section-14.7

Which should be handled accordingly to rfc3986: https://datatracker.ietf.org/doc/html/rfc4918#section-3
So percent encoding is perfectly fine and should be used if needed.

perry-mitchell:
Closing as it looks like the consensus is that this isn't needed. Thanks to all involved.

PR #324, opened 2022-10-19
fix: Fix missing leading slash causing extra getDirectoryContents entries

Cobertos:
If you don't have a leading slash passed to `getDirectoryContents`, it will cause it to not properly filter out the current directory from the response. This PR fixes that.

This is the actual cause of the issue I was having in the `webdav-fs` repo. https://github.com/perry-mitchell/webdav-fs/issues/84


perry-mitchell:
For some reason the tests won't run on PRs from forks, so I had to merge this to a temporary branch to run them.

perry-mitchell:
Worked! Thanks @Cobertos for this fix.. Much appreciated :pray:

PR #341, opened 2023-05-13
Preserve stringified numbers with leading zeros

Pytal:
Enable the `parseTrueNumberOnly` option from `fast-xml-parser`, see [docs](https://github.com/NaturalIntelligence/fast-xml-parser/blob/master/docs/v3/docs.md), to preserve the leading zeros of numbers

### Before
`"00123"` is parsed to `123`

### After
`"00123"` is parsed to `"00123"`


perry-mitchell:
This is great! I can merge later today. A test would go great with this but I can also try to add one when I get time.

PR #350, opened 2023-08-03
fix: Update dependencies and audit

susnux:
* `nodemon` before version 3 introduces security issues reported by npm audit, so updated to current v3
* `minimatch` v8 or newer require at least Node 14.17 so no compatible without a breaking change


perry-mitchell:
> minimatch v8 or newer require at least Node 14.17 so no compatible without a breaking change

But the minimum node version here is v14 :)

image

PR #356, opened 2023-08-23
fix: The `exports` field should use wildcard exports instead of folder mapping

### LINKED ISSUE #355: Can not be imported with newer node versions
The package can not be imported by some build tools and newer node versions, because the `exports` field of the package.json contain a folder mapping:

https://github.com/perry-mitchell/webdav-client/blob/630168b8183b2c89b71e38149ff502366ddee6d1/package.json#L11

This is deprecated since Node 14 and removed in Node 17. It should be replaced with:
```
"./dist/*": "./dist/*"
```

susnux:
* Resolved #355

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
       "default": "./dist/node/index.js"
     },
     "./web": "./dist/web/index.js",
-    "./dist/": "./dist/"
+    "./dist/*": "./dist/*"
   },
   "react-native": "./dist/web/index.js",
   "type": "module",


susnux:
@perry-mitchell would be great if this could be fixed as it requires to patch the package on every install otherwise :sweat_smile:

perry-mitchell (approved):

skjnldsv:
:tada: Thank you as always @perry-mitchell !! :hugs:

perry-mitchell:
Planning to do releases for all this tonight!

PR #360, opened 2023-11-24
fix: Relative filename

Pytal:
Fix incorrect filename due to the XML encoded href property being compared to a possibly non-encoded base url

### Before
```jsonc
"filename": "/../aA0 _.@-'"
```

### After
```jsonc
"filename": "/"
```


susnux (approved):

perry-mitchell (approved):

perry-mitchell:
Released in 5.3.1 - Thanks for this!

PR #362, opened 2024-01-11
Add partialUpdateFileContents method

### LINKED ISSUE #303: Is it possible to append a new line to file on webdav server?
Hi , thanks for your cool tool first , Is it possible to append a new line to file on webdav server? 
for now , I have to fetch remote file first, then append new line, then overwrite the remote file again.
It's better if there's a `append line` on put method directly
Thanks

ArcticLampyrid:
This pull request adds a new method called `partialUpdateFileContents` to the WebDAVClient class. This method allows for updating a remote file with a partial update, which is useful for making changes to a file without having to download and re-upload the entire file.

Moreover, it can be used to implement chunked file uploading.

This will close #303.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,26 @@ await client.putFileContents("/my/file.txt", str);
 
 _`options` extends [method options](#method-options)._
 
+#### partialUpdateFileContents
+
+Update a remote file with a partial update. This method is useful for updating a file without having to download and re-upload the entire file.
+
+**_Note that this method is not standardised and may not be supported by all servers._** To use this feature, one of the following must be met:
+ * WebDav is served by Apache with the [mod_dav](https://httpd.apache.org/docs/2.4/mod/mod_dav.html) module
+ * The server supports Sabredav PartialUpdate Extension (https://sabre.io/dav/http-patch/)
+
+```typescript
+(filePath: string, start: number, end: number, data: string | BufferLike | Stream.Readable, options?: WebDAVMethodOptions)=> Promise<void>
+```
+
+| Argument          | Required  | Description                                   |
+|-------------------|-----------|-----------------------------------------------|
+| `filePath`        | Yes       | File to update.                               |
+| `start`           | Yes       | Start byte position. (inclusive)              |
+| `end`             | Yes       | End byte position.   (inclusive)              |
+| `data`            | Yes       | The data to write. Can be a string, buffer or a readable stream. |
+| `options`         | No        | Configuration options.                        |
+
 #### search
 
 Perform a WebDAV search as per [rfc5323](https://www.ietf.org/rfc/rfc5323.html).
--- a/source/factory.ts
+++ b/source/factory.ts
@@ -16,7 +16,7 @@ import { getSearch } from "./operations/search.js";
 import { moveFile } from "./operations/moveFile.js";
 import { getFileUploadLink, putFileContents } from "./operations/putFileContents.js";
 import { partialUpdateFileContents } from "./operations/partialUpdateFileContents.js";
-import { getDavCompliance } from "./operations/getDavCompliance.js";
+import { getDAVCompliance } from "./operations/getDAVCompliance.js";
 import {
     AuthType,
     BufferLike,
@@ -115,7 +115,7 @@ export function createClient(remoteURL: string, options: WebDAVClientOptions = {
             data: string | BufferLike | Stream.Readable,
             options?: WebDAVMethodOptions
         ) => partialUpdateFileContents(context, filePath, start, end, data, options),
-        getDavCompliance: (path: string) => getDavCompliance(context, path),
+        getDAVCompliance: (path: string) => getDAVCompliance(context, path),
         search: (path: string, options?: SearchOptions) => getSearch(context, path, options),
         setHeaders: (headers: Headers) => {
             context.headers = Object.assign({}, headers);
--- a/source/factory.ts
+++ b/source/factory.ts
@@ -15,6 +15,8 @@ import { getStat } from "./operations/stat.js";
 import { getSearch } from "./operations/search.js";
 import { moveFile } from "./operations/moveFile.js";
 import { getFileUploadLink, putFileContents } from "./operations/putFileContents.js";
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -554,6 +554,26 @@ await client.putFileContents("/my/file.txt", str);
 
 _`options` extends [method options](#method-options)._
 
+#### partialUpdateFileContents
+
+Update a remote file with a partial update. This method is useful for updating a file without having to download and re-upload the entire file.
+
+**_Note that this method is not standardised and may not be supported by all servers._** To use this feature, one of the following must be met:
+ * WebDav is served by Apache with the [mod_dav](https://httpd.apache.org/docs/2.4/mod/mod_dav.html) module
```
n-peugnet, line 562: Out of curiosity, where did you find evidence that Apache's mod_dav supports this feature ?

ArcticLampyrid: Actually, from the sources of Apache:
https://github.com/apache/httpd/blob/c9028584a89a761e2804b596ce2c9f961f5b6fb8/modules/dav/main/mod_dav.c#L1094-L1108

And some discussions on other WebDAV related projects. (eg. https://github.com/miquels/webdav-handler-rs/blob/master/doc/Apache-PUT-with-Content-Range.md)

source/operations/getDavCompliance.ts
```diff
@@ -0,0 +1,39 @@
+import { joinURL } from "../tools/url.js";
+import { encodePath } from "../tools/path.js";
+import { request, prepareRequestOptions } from "../request.js";
+import { handleResponseCode } from "../response.js";
+import {
+    WebDAVMethodOptions,
+    WebDAVClientContext,
+    WebDAVClientError,
+    DavCompliance
```
perry-mitchell, line 9: Please keep alphabetical

source/operations/getDavCompliance.ts
```diff
@@ -0,0 +1,39 @@
...
+} from "../types.js";
+
+export async function getDavCompliance(
```
perry-mitchell, line 12: Can you rename this `getDAVCompliance`?

source/factory.ts
```diff
@@ -106,6 +108,14 @@ export function createClient(remoteURL: string, options: WebDAVClientOptions = {
...
             options?: PutFileContentsOptions
         ) => putFileContents(context, filename, data, options),
+        partialUpdateFileContents: (
+            filePath: string,
+            start: number,
+            end: number,
+            data: string | BufferLike | Stream.Readable,
+            options?: WebDAVMethodOptions
+        ) => partialUpdateFileContents(context, filePath, start, end, data, options),
+        getDavCompliance: (path: string) => getDavCompliance(context, path),
```
perry-mitchell, line 118: And `getDAVCompliance` here too..

source/operations/partialUpdateFileContents.ts
```diff
@@ -0,0 +1,136 @@
+import { Layerr } from "layerr";
+import Stream from "stream";
```
perry-mitchell, line 2: Please keep at the top, and use `node:stream`

perry-mitchell: Also, you could just import `{ Readable }` instead of `stream`

source/operations/getDavCompliance.ts
```diff
@@ -0,0 +1,39 @@
...
+        },
+        context,
+        options
+    );
+    const response = await request(requestOptions);
+    try {
+        handleResponseCode(context, response);
+    } catch (err) {
+        const error = err as WebDAVClientError;
+        throw error;
```
perry-mitchell, line 30: Why catch if we're just throwing again? Maybe wrap in `Layerr` to give more insight/context? Improving the error message with Layerr would be fine.

source/operations/partialUpdateFileContents.ts
```diff
@@ -0,0 +1,136 @@
...
+        },
+        context,
+        options
+    );
+    const response = await request(requestOptions);
+    try {
+        handleResponseCode(context, response);
+    } catch (err) {
+        const error = err as WebDAVClientError;
+        throw error;
```
perry-mitchell, line 134: Same here. Throwing a wrapped error with `Layerr` would be more useful here.

source/types.ts
```diff
@@ -133,6 +135,11 @@ export interface SearchResult {
     results: FileStat[];
 }
 
+export interface DavCompliance {
```
perry-mitchell, line 138: Please rename `DAVCompliance`.

perry-mitchell (changes requested):
Some minor items..

perry-mitchell:
Awesome PR here! Thanks @ArcticLampyrid. The DAV compliance fetcher is very nice and I imagine will come in super useful in the future.

Some minor items in the PR, like naming etc.

source/operations/getDavCompliance.ts
ArcticLampyrid: It seems that there is not much contextual information here.

I've just removed `try ... catch ...`

source/operations/partialUpdateFileContents.ts
ArcticLampyrid: Same as above.

perry-mitchell (approved):

perry-mitchell:
@ArcticLampyrid Tests failing due to drop in coverage. Would you mind seeing if there are some lines added in this PR that are not covered by tests? And if so, adding coverage for them?

ArcticLampyrid:
The decrease in test coverage is expected. This PR adds partialUpdateFileContents which relies on a non-standard extension, and the webdav-server used during testing does not yet support this extension. In my opinion, there is no simple way to add complete tests. The current unit tests only cover the "not support" scenario.

perry-mitchell:
Ok. Would you mind then dropping the coverage amounts to match the new values? I'll merge after that and once the tests pass.

ArcticLampyrid:
> Would you mind then dropping the coverage amounts to match the new values?

Of course, I'd be happy to adjust the coverage amounts to align with the new values.

perry-mitchell:
Sorry this took so long to move. Merging.

PR #364, opened 2024-02-04
fix: Do not double decode XML entities in filenames

### LINKED ISSUE #363: XML entities in filenames are not handled correctly
There is a problem with XML entities within filenames:
Given file `Hello &amp;.txt` and its URL encoding `Hello%20%26amp%3b.txt`.

The *PROPFIND* result looks like this on Apache and SabreDAV:
```xml
<d:response>
    <d:href>/Hello%20%26amp%3b.txt</d:href>
```

The `getDirectoryContents` now decodes the URL encoding (correctly):

https://github.com/perry-mitchell/webdav-client/blob/e41a805af2ed102ce581058c411b69e5b7d94a58/source/operations/directoryContents.ts#L78

So the filename now is `Hello &amp;.txt` like it is supposed to be. But `prepareFileFromProps` now also decodes HTML entities which breaks the filename as it is now `Hello &.txt` which is not the correct filename.

https://github.com/perry-mitchell/webdav-client/blob/e41a805af2ed102ce581058c411b69e5b7d94a58/source/tools/dav.ts#L132

susnux:
* Resolves #363 

The `href` property is already decoded by `parseXML` meaning all XML entities like `&amp;` were decoded. The href just needs to be url decoded.

This prevents issues were files like `a &amp; b.txt` were returned with the wrong name `a & b.txt`

Ref: https://datatracker.ietf.org/doc/html/rfc4918#section-8.3

### DIFF
--- a/source/operations/directoryContents.ts
+++ b/source/operations/directoryContents.ts
@@ -66,7 +66,7 @@ function getDirectoryFiles(
 
     // Map all items to a consistent output structure (results)
     const nodes = responseItems.map(item => {
-        // HREF is the file path (in full)
+        // HREF is the file path (in full) - The href is already XML entities decoded (e.g. foo&amp;bar is reverted to foo&bar)
         const href = normaliseHREF(item.href);
         // Each item should contain a stat object
         const {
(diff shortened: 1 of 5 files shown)


perry-mitchell:
Thanks @susnux ! Everything looks good here so I'll merge.

perry-mitchell:
@susnux Released in 5.3.2

PR #376, opened 2024-07-02
fix: also export types for node next module resolution

susnux:
When using node next / node16 module resolution (`exports` field), then Typescript will no longer check `types` field in package json but use `types` from the `exports` field that is currently being used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
     ".": {
       "browser": "./dist/web/index.js",
       "react-native": "./dist/react-native/index.js",
+      "types": "./dist/node/index.d.ts",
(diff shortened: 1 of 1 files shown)


perry-mitchell (approved):
Looks good, thanks!

perry-mitchell:
Not sure why the tests are failing. I'll check it a bit later.

susnux:
@perry-mitchell Every run fails on a other test, it seems like the `await` does not resolve.
So maybe this is a issue with the webdav server that times out?

But I can not think of any way this is related to this PR.

perry-mitchell (approved):

PR #378, opened 2024-07-24
fix(types): correct types for quota bytes

susnux:
As per RFC the value shall be octets representing the quota, in all cases I know the value will be parsed to `number` by the XML parser. (tested e.g. with Sabre / Nextcloud).

While `any` allowed everything, the type `string` for *quota-used-bytes* was wrong as mostly a number is returned (but there might be rare cases of string, like "unlimited" (even if the RFC does not mention this), so lets keep the `string` type in addition.

### DIFF
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -237,7 +237,7 @@ export function parseSearch(result: DAVResult, searchArbiter: string, isDetailed
  * @returns The value in bytes, or another indicator
  */
 export function translateDiskSpace(value: string | number): DiskQuotaAvailable {
-    switch (value.toString()) {
+    switch (String(value)) {
         case "-3":
             return "unlimited";
         case "-2":
@@ -246,6 +246,6 @@ export function translateDiskSpace(value: string | number): DiskQuotaAvailable {
             // -1 is non-computed
             return "unknown";
         default:
-            return parseInt(value as string, 10);
+            return parseInt(String(value), 10);
     }
 }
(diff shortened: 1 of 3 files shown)


source/tools/dav.ts
```diff
@@ -233,7 +233,8 @@ export function parseSearch(result: DAVResult, searchArbiter: string, isDetailed
  * @returns The value in bytes, or another indicator
  */
 export function translateDiskSpace(value: string | number): DiskQuotaAvailable {
-    switch (value.toString()) {
+    value = String(value);
```
perry-mitchell, line 236: Let's not re-assign function args - could you please make this a new constant?

perry-mitchell: Or even better, just put the `String` call in the switch arg.

perry-mitchell (changes requested):

source/tools/dav.ts
```diff
@@ -233,7 +233,8 @@ export function parseSearch(result: DAVResult, searchArbiter: string, isDetailed
...
+    switch (value) {
```
susnux, line 237:
```suggestion
    switch (String(value)) {
```

source/tools/dav.ts
```diff
@@ -242,6 +243,6 @@ export function translateDiskSpace(value: string | number): DiskQuotaAvailable {
             // -1 is non-computed
             return "unknown";
         default:
-            return parseInt(value as string, 10);
+            return parseInt(value, 10);
```
susnux, line 246:
```suggestion
            return parseInt(String(value), 10);
```

source/tools/dav.ts
susnux: makes sense, adjusted the code!

perry-mitchell (approved):

perry-mitchell:
Thanks @susnux !

PR #380, opened 2024-07-24
fix(dav): The `displayname` prop should always be a string

susnux:
In case e.g. a directory is called `1` the XML parser will convert the display name to a number. This is against the types and not expected behavior.

### DIFF
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -146,6 +146,10 @@ export function prepareFileFromProps(
         stat.mime = mimeType && typeof mimeType === "string" ? mimeType.split(";")[0] : "";
     }
     if (isDetailed) {
+        // The XML parser tries to interpret values, but the display name is required to be a string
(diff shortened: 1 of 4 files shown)


source/tools/dav.ts
```diff
@@ -146,6 +146,10 @@ export function prepareFileFromProps(
         stat.mime = mimeType && typeof mimeType === "string" ? mimeType.split(";")[0] : "";
     }
     if (isDetailed) {
+        // The XML parser tries to interpret values, but the display name is required to be a string
+        if (props.displayname !== undefined) {
```
perry-mitchell, line 150: Could you please make this `typeof props.displayname !== "undefined"`?

perry-mitchell (changes requested):
~~Just one item! Thanks for this..~~

perry-mitchell (approved):

PR #384, opened 2024-08-27
fix: Ensure that the `displayname` property is not parsed

susnux:
If you have nodes with number-like names those values might be corrupted. While integer like names would be casted back to string, float-like names would be corrupted.
E.g. `2024.10` will be parsed to number `2024.1` thus loosing the trailing 0.

---

@perry-mitchell I think there might be even more places, especially with custom properties, where the parsing will lead to unexpected behavior. So maybe disable parsing in general and let users parse it?
Or at least expose the preprocessors, so users can decide which prop to parse.

But I think that would be a breaking change so not in 5.x. So this PR only fixes the displayname.

### DIFF
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -27,6 +27,13 @@ function getParser(): XMLParser {
         numberParseOptions: {
             hex: true,
             leadingZeros: false
+        },
+        tagValueProcessor(tagName, tagValue, jPath) {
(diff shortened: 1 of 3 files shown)


perry-mitchell (approved):

susnux:
Would it be possible to create a new patch release for this @perry-mitchell ? :)

PR #391, opened 2024-12-09
feat: support attributes parsing

### LINKED ISSUE #388: Support xml attributes
## Xml example
```xml
<?xml version="1.0"?>
<d:multistatus
	xmlns:d="DAV:"
	xmlns:s="http://sabredav.org/ns"
	xmlns:oc="http://owncloud.org/ns"
	xmlns:nc="http://nextcloud.org/ns">
<d:response>
		<d:href>/remote.php/dav/files/admin/benjamin-lehman-xnTcEJm5IZw-unsplash.jpg</d:href>
		<d:propstat>
			<d:prop>
				<nc:system-tags>
					<nc:system-tag oc:can-assign="true" oc:id="321" oc:user-assignable="true" oc:user-visible="true">Test</nc:system-tag>
				</nc:system-tags>
			</d:prop>
			<d:status>HTTP/1.1 200 OK</d:status>
		</d:propstat>
	</d:response>
</d:multistatus>
```

## Issue
Currently, the `fast-xml-parser` ignore attributes.
A quick test, passing [the following options](https://github.com/perry-mitchell/webdav-client/blob/193268946a608708733a15612caaab13f93b156c/source/tools/dav.ts#L25-L30) would allow us to retrive the attributes like so
```js
        attributeNamePrefix: "",
        ignoreAttributes: false,
        textNodeName: "text",
```
image

I think adding a proper method to process those would be best, so we can keep the namespaces and avoid conflicts?

skjnldsv:
Fix #388 

What do you think of that [perry-mitchell](https://github.com/perry-mitchell/) :)

### DIFF
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -23,10 +23,21 @@ enum PropertyType {
 
 function getParser(): XMLParser {
     return new XMLParser({
+        allowBooleanAttributes: true,
+        attributeNamePrefix: "",
+        textNodeName: "text",
+        ignoreAttributes: false,
         removeNSPrefix: true,
         numberParseOptions: {
             hex: true,
             leadingZeros: false
+        },
+        attributeValueProcessor(attrName, attrValue, jPath) {
+            // handle boolean attributes
+            if (attrValue === "true" || attrValue === "false") {
+                return attrValue === "true";
+            }
+            return attrValue;
         },
         tagValueProcessor(tagName, tagValue, jPath) {
             if (jPath.endsWith("propstat.prop.displayname")) {
--- a/test/node/operations/stat.spec.ts
+++ b/test/node/operations/stat.spec.ts
@@ -130,6 +130,38 @@ describe("stat", function () {
         });
     });
 
+    it("correctly parses the attributes property", async function () {
+        returnFakeResponse(
+            readFileSync(
+                new URL("../../responses/propfind-attributes.xml", import.meta.url)
+            ).toString()
+        );
+
+        await this.client.stat("/foo/", { details: true }).then(function (result) {
+            expect(result.data).to.have.property("props").that.is.an("object");
+            expect(result.data.props).to.have.property("system-tags").that.is.an("object");
+            expect(result.data.props["system-tags"]["system-tag"])
+                .to.be.an("array")
+                .and.have.lengthOf(2);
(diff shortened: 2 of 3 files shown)


perry-mitchell (approved):
Love it! Sorry it took forever to merge..

susnux:
This is potentially doing unexpected things, as this similar to other parsing, cannot be disabled.
Meaning if you expect a string and its randomly named "true" you now get a boolean instead of a string.

This could lead to similar issues like fixed with:
https://github.com/perry-mitchell/webdav-client/pull/384

I think the proper way of handling all this parsing would be to only parse attributes where the type is specified by the WebDAV RFCs and for all other make it opt-in. Users normally know it best what type their properties are.

skjnldsv:
> This could lead to similar issues like fixed with:
> #384

Iirc, the current behaviour is to eval booleans.
The PR you linked #384, basically added an exception for the displayName only, right?

Just to be clear, this PR #391, doesn't really change the way this lib is behaving as far as property evaluation goes? Maybe I'm misunderstanding, what are you suggesting we do @susnux :)

susnux:
@skjnldsv the problem is you added new behavior probably causing unexpected behavior because it always thinks "true" and "false" are boolean independent from the attribute they are set on.
e.g. `id="true"` would be `id: true` even if you expect an ID to be a string.

Also you added no attribute prefix, this will cause attributes to clash with same named child props.
E.g.:

`<foo bar><bar>baz</bar></foo>` will result in either:
`foo: { bar: true }` or `foo: { bar: { text: 'baz' } }` not sure if randomly or if the parser has  a preference but it will hide the other attribute / prop.

susnux:
BTW this was also a breaking change:

given response:
```xml
<prop>
    <some-prop attribute>value</some-prop>
</prop>
```

Before it resulted in:
```js
{
    prop: {
        'some-prop': 'value'
    }
}
```

So only the attribute was missing, but now the attribute is kept but the value is now on the `text` attribute:

```js
{
    prop: {
        'some-prop': {
            text: 'value',
            attribute: true,
    }
}
```

susnux:
The solution I see here is to let the library user decide when to parse or not:
https://github.com/perry-mitchell/webdav-client/pull/395

skjnldsv:
> The solution I see here is to let the library user decide when to parse or not: #395

Good from me :)

PR #395, opened 2025-02-13
feat: Allow to register prop and attribute parsers

susnux:
This consists of two changes:

1. A fix for https://github.com/perry-mitchell/webdav-client/pull/391#issuecomment-2649178022
   **Allow to configure a tag prefix.**
   Without a prefix there can be name clashes between attributes and child
    tags. For example `<tag id="A"><id>B</id></tag>` will not be parsed to
    contain both the tag `id` and the attribute `id` but one will overwrite
    the other.
2. Allow to register custom parsers for attributes and props, so we do not always do automatic numeric parsing leading to issues like we had with the `displayname`, see the examples in the added tests on how this could be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,7 +717,7 @@ function jsonPropParser(jPath: string, value: string) {
         return JSON.parse(value)
     }
     // Apply default parsing otherwise
-    return value
+    return value;
 }
 await client.registerTagParser(jsonPropParser);
 ```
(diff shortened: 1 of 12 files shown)


perry-mitchell (changes requested):
Couple of small items..

perry-mitchell:
Thank you @susnux for this! Definitely saving me from pushing the last changes without checking 😅 

I suppose they wouldn't likely affect standard use cases, but more custom webdav calls right? Where end users are expecting to see more of the "raw" dav output?

susnux:
@perry-mitchell thank you for the documentation fixes!

> I suppose they wouldn't likely affect standard use cases, but more custom webdav calls right? Where end users are expecting to see more of the "raw" dav output?

Yes exactly. 

> Can you confirm that this reverts the behaviour to what it was before 5.8.0 release?

yes + the new feature of attributes parsing. No not breaking anymore.

Issue #210, opened 2020-06-09
How to handle namespace in props fetched by stat() ?


yoh1496:
Hi, I'm implementing PROPPATCH method ( related to #206 ).

I'm writing test codes. ( I'm referring to to [webdav-server test code](https://github.com/OpenMarshal/npm-WebDAV-Server/blob/master/test/v2/tests.ts/properties/.test.ts).)

Some properties have namespace such as
- `<namespace:testcustom>Value</namespace:testcustom>` 
- `<x:testcustom2 xmlns:x="namespace:">Value</x:testcustom2>`

but stat function with `{ details: true }` options returns only tagName without namespaces.

Please let me know how to handle namespaces in result XML ?

FYI, I think combination of `xpath` and `xmldom` is useful for handling namespace if `stat` would return raw XML data.

Thanks.

perry-mitchell:
I actually forced it to strip them, as they're not useful for the majority of basic DAV requests. Why would you need the namespaces?

yoh1496:
Thank you for responding to me.

According to RFC4918, property names consist of [namespace name], [local name], [attributes], and [children].

Actually, the `webdav-server`'s test has code for testing that you can correctly assign properties named with namespace. 

And the reason I need namespaces are that I encountered usecase of using properties with namespace. In that case, `PROPFIND` result stripped namespaces cannot be used. ( developers have to use `PROPFIND` method manually. )

If you need more info, please ask.

Thanks

perry-mitchell:
For this I think I would need either:

 * An example service where namespaces are relevant to this webdav client (for file/directory manipulation)
 * Someone to contribute correct namespace handling code, with test cases against webdav-server

Happy to leave this open as an issue.

PR #410, opened 2026-05-21
feat: opt-in Clark-notation prop keys for namespace disambiguation

dschmidt:
Related to #210, though not a full fix for the high-level `stat()` shape it asks about.

## Motivation

I'm building a generic extension metadata mechanism in OpenCloud. Third-party extensions can attach typed metadata to files, and the natural mapping for cross-protocol access turned out to be one custom XML namespace per extension. Multiple extensions can then reuse the same local property name without colliding, because the namespace URI is what disambiguates them. RFC 4918 supports this cleanly, and on the server side it falls out for free.

On the client side it doesn't. Today `parseXML` strips namespace prefixes unconditionally, so two properties with the same local name from different namespaces (e.g. `<value xmlns="proj-ns">` and `<value xmlns="rev-ns">`) collapse to a single `value` key (with values merged into an array) and lose their namespace origin entirely. Real-world servers add a second twist: Go's `encoding/xml` and similar marshalers serialise the standard DAV/oc properties with prefixes (`<d:getlastmodified>`) but custom-namespace properties with **inline default xmlns** on the element (`<value xmlns="...">`), because they have no prefix registered for the custom namespace. The client has to deal with both serialisation styles in the same response.

## What I originally tried

My first attempt at this PR was a simple opt-in that just told `parseXML` to keep namespace prefixes on tags (essentially `removeNSPrefix: false`). That works for servers that prefix-serialise custom namespaces, but for the inline-`xmlns` style it just leaked fast-xml-parser's raw object shape into the result (`{ "text": "...", "@xmlns": "..." }` per element, arrays when collisions occur, and so on). That made the option's output awkward to reason about and inconsistent across server serialisation styles. I'd be shipping a half-feature whose consumers would commit to an unpleasant shape, and any "real" fix later would mean a second parallel opt-in.

So the actual feature is Clark notation: one canonical key per property regardless of how the server expressed the namespace.

## What this PR does

Adds an optional `clarkNotationProps` field to `WebDAVParsingContext`, defaulting to `false`. When set to `true`, every property key inside each `propstat.prop` is rewritten to Clark notation `{namespaceURI}localName`. The structural shape of `DAVResult` is unaffected: `multistatus`, `response`, `propstat`, `prop`, `status`, `href` all keep their bare names, and the existing `normaliseResult` runs unchanged.

Downstream helpers like `prepareFileFromProps`, `parseStat` and `parseSearch` assume bare prop keys and are intentionally left untouched; consumers that opt in handle the Clark-notation keys on their side, which is a single property access if they know the namespace they're after.

## Implementation notes

The implementation isn't completely trivial and I'd really appreciate review feedback. A single-pass tree walker runs after `fast-xml-parser` parses with `removeNSPrefix: false` and does two things in one traversal:

1. Renames any prefixed structural keys back to their bare local names so `normaliseResult` can do its job.
2. Inside each `<prop>`, walks the children, resolves each one's namespace (from the xmlns scope of the multistatus element, or from an inline `xmlns="..."` on the element itself), unwraps the `{ text, "@xmlns" }` object shape, and emits a single Clark-notation key per property. Same-local-name collisions in different namespaces become distinct keys.

Edge cases:
- Unknown prefixes (used but never declared) fall back to the null namespace; that yields a bare local-name key instead of throwing.
- Multiple elements with the same Clark key (genuine collisions in the same namespace) are kept as an array under that key.
- The walk extends the prefix-to-URI scope map only when a node actually declares xmlns, to avoid per-node allocations.
- String references for the xmlns/text attribute keys are computed once at the start of the walk so V8 can inline-cache the hot lookups.

I tried to keep the cost as low as I could by benchmarking alternative walker designs (V8 micro-optimisations, algorithmic restructure with direct navigation of the envelope, and parser-cooperative approaches using fast-xml-parser hooks). The full benchmark suite landed within noise of the current shape on end-to-end `parseXML`, since the parser dominates the cost and the walker is a small slice on top. I left the walker as a clean generic traversal rather than commit a more complex variant for a wash result. Very open to suggestions if you see a cheaper structural pattern.

## Performance

Benchmarked against the current default and a discarded two-pass alternative. Each benchmark fixture is a synthetic multistatus response: one `<d:response>` per file with a single `<d:propstat>` (status 200) containing 8 standard properties (`d:getlastmodified`, `d:getcontentlength`, `d:getetag`, `d:getcontenttype`, `d:resourcetype`, `oc:permissions`, `oc:id`, `oc:fileid`) plus additional extension properties from two distinct custom namespaces with inline default-xmlns serialisation (the realistic Sabre / Go-`encoding/xml` shape), rotating, up to the configured prop count per file.

| scenario | size | default (strip) | clarkNotationProps | overhead |
|---|---:|---:|---:|---:|
| 10 files / 10 props | 8.5 KB | 0.38 ms | 0.41 ms | +6% |
| 100 files / 15 props | 134 KB | 4.96 ms | 5.96 ms | +20% |
| 1k files / 20 props | 1.8 MB | 69 ms | 84 ms | +21% |
| 5k files / 30 props | 14 MB | 514 ms | 640 ms | +24% |
| 10k files / 50 props | 49 MB | 1674 ms | 2125 ms | +27% |
| 20k files / 50 props | 99 MB | 3343 ms | 4279 ms | +28% |

Overhead settles around 25% on realistic-and-large inputs, paid only by consumers that opt in.

## Tests

Three cases under `parseXML > clarkNotationProps`: documents today's collapse-on-collision behaviour (lossy array of values), verifies that the Clark-notation rewrite yields distinct canonical keys covering both prefix-resolved (DAV/oc) and inline-xmlns (extension) properties, and a malformed-input case where an undeclared prefix falls back to the null namespace rather than crashing.

### DIFF
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -187,11 +187,6 @@ const STRUCTURAL_KEYS = new Set([
     "responsedescription"
 ]);
 
-function localName(key: string): string {
-    const colonIdx = key.indexOf(":");
-    return colonIdx === -1 ? key : key.slice(colonIdx + 1);
-}
-
 function isPlainObject(value: unknown): value is Record<string, unknown> {
     return typeof value === "object" && value !== null && !Array.isArray(value);
 }
@@ -204,19 +199,15 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
  *  2. Rewrite every key inside each `<prop>` to Clark notation
  *     `{namespaceURI}localName`, resolving the namespace from the xmlns
  *     scope of the surrounding multistatus and from any inline
- *     `xmlns="..."` declared on the property element itself. Wrapped values
- *     of the form `{ "@xmlns": ..., text: "..." }` are unwrapped to just
- *     their text content.
- *
- * Pre-computes the xmlns attribute keys once and extends the prefix scope
- * map only when a node actually declares xmlns, so the walk stays close to
- * a plain tree traversal in cost.
+ *     `xmlns="..."` declared on the property element itself. Wrapped
+ *     simple-text elements of the form `{ "@xmlns": ..., text: "..." }`
+ *     are unwrapped to their text content. Property values with their own
+ *     nested element children are returned as-is (the keys of those
+ *     grandchildren are not rewritten to Clark notation).
  */
 function applyClarkNotation(root: unknown, attrPrefix: string): void {
     if (!isPlainObject(root) && !Array.isArray(root)) return;
 
-    // Stable string references for the duration of the walk; lets V8
-    // inline-cache the property accesses against the same key shapes.
     const xmlnsKey = `${attrPrefix}xmlns`;
     const xmlnsPrefixedKey = `${xmlnsKey}:`;
     const xmlnsPrefixedKeyLen = xmlnsPrefixedKey.length;
@@ -249,26 +240,19 @@ function applyClarkNotation(root: unknown, attrPrefix: string): void {
         return scope ?? parent;
     }
 
+    // Unwrap the `{ "@xmlns": ..., text: "..." }` shape that fast-xml-parser
+    // produces for elements with both an xmlns attribute and text content.
+    // Complex elements with their own nested children are returned as-is:
+    // their keys are NOT rewritten to Clark notation, since the walker only
+    // resolves namespaces at the `<prop>` child level.
     function unwrapXmlnsWrappedValue(raw: unknown): unknown {
         if (!isPlainObject(raw)) return raw;
         const text = raw[TEXT_KEY];
-        // If the only non-text keys are xmlns attrs, the element is a simple
-        // text node with namespace metadata: extract the text directly.
-        let onlyXmlnsAndText = true;
         for (const k of Object.keys(raw)) {
             if (k === TEXT_KEY) continue;
-            if (!isXmlnsAttr(k)) {
-                onlyXmlnsAndText = false;
-                break;
-            }
+            if (!isXmlnsAttr(k)) return raw;
         }
-        if (onlyXmlnsAndText && text !== undefined) return text;
-        // Complex element: clone with xmlns attrs stripped, keep child content.
-        const out: Record<string, unknown> = {};
-        for (const k of Object.keys(raw)) {
-            if (!isXmlnsAttr(k)) out[k] = raw[k];
-        }
-        return out;
+        return text !== undefined ? text : raw;
     }
 
     function emitClarkForChild(
@@ -335,7 +319,8 @@ function applyClarkNotation(root: unknown, attrPrefix: string): void {
         const childScope = extendScope(node, scope);
 
         for (const key of Object.keys(node)) {
-            const ln = localName(key);
+            const colonIdx = key.indexOf(":");
+            const ln = colonIdx === -1 ? key : key.slice(colonIdx + 1);
             const value = node[key];
 
             // Rename structural keys to bare local name.
--- a/source/tools/dav.ts
+++ b/source/tools/dav.ts
@@ -23,13 +23,26 @@ enum PropertyType {
     Original = "original"
 }
 
-function toJPathString(
-    jPath: string | { toString: (sep?: string, includeNS?: boolean) => string }
-): string {
+type JPath = string | { toArray: () => string[] };
+
+function stripPrefix(name: string): string {
+    const colonIdx = name.indexOf(":");
+    return colonIdx === -1 ? name : name.slice(colonIdx + 1);
+}
+
+function isXmlnsAttribute(name: string): boolean {
+    return name === "xmlns" || name.startsWith("xmlns:");
+}
+
+// The parser only keeps namespace prefixes on tags in Clark mode, so without
+// this the jPaths would be `d:propstat.d:prop.d:displayname` there and bare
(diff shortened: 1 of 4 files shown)


dschmidt:
I just realised OpenCloud uses xmlns Attributes on the props - figuring out what this means for this PR :)

dschmidt:
Feature got a bit bigger than anticipated, but I'm pretty satisfied with the outcome. Appreciate any feedback!

dschmidt:
@perry-mitchell any thoughts?

perry-mitchell:
Hi @dschmidt - sorry I haven't gotten to this, just checking it now.. Thanks for taking the effort!

dschmidt:
Great, I'm still here :) 

Let me know what you think and I'm happy to iterate on this. I would really still like to see this upstream.

source/tools/dav.ts
```diff
@@ -43,7 +44,7 @@ function getParser({
         attributeNamePrefix,
         textNodeName: "text",
         ignoreAttributes: false,
-        removeNSPrefix: true,
+        removeNSPrefix: !clarkNotationProps,
```
perry-mitchell, line 60: `tagParsers` / `attributeParsers` receive prefixed jPaths when clark mode is on (`displaynameTagParser`'s `endsWith("propstat.prop.displayname")` never matches). Maybe document or normalise the path before invoking processors?

source/types.ts
```diff
@@ -408,6 +414,7 @@ export interface WebDAVEntityDecoderOptions {
 export interface WebDAVParsingContext {
     attributeNamePrefix?: string;
     attributeParsers: WebDAVAttributeParser[];
+    clarkNotationProps?: boolean;
```
perry-mitchell, line 434: Not plumbed through `WebDAVClientOptions` / `factory.ts`, so only reachable via exported parseXML. Perhaps worth stating explicitly in the JSDoc that it's a low-level escape hatch, and that high-level methods would break if ever wired in?

source/tools/dav.ts
```diff
@@ -165,9 +166,207 @@ function normaliseResult(result: DAVResultRaw): DAVResult {
...
+            if (existing === undefined) {
+                out[key] = value;
+            } else if (Array.isArray(existing)) {
+                existing.push(value);
+            } else {
+                out[key] = [existing, value];
+            }
+        };
+
+        if (Array.isArray(rawValue)) {
```
perry-mitchell, line 321: Same-Clark-key collision -> array branch is untested. Perhaps a test with two <value> elements in the same namespace (e.g. alongside the existing cross-namespace collapse test) would be best?

perry-mitchell (changes requested):

source/tools/dav.ts
dschmidt: Good catch, that was a real bug. Normalised now: the jPath gets its prefixes stripped per segment before the processors run, so parsers see the same paths in both modes and displaynameTagParser fires again. While in there I also skipped the xmlns declarations, they only survive in clark mode and were showing up as extra attributeParser calls (and a parser could rewrite the URI and corrupt the clark keys that way). Test asserts both modes hand out identical paths.

source/types.ts
dschmidt: Done, the JSDoc now says it is a low-level parseXML escape hatch and why it is not on WebDAVClientOptions (stat/getDirectoryContents/search/getQuota all read bare prop keys). Also added a short README section so the option is at least discoverable.

source/tools/dav.ts
dschmidt: Added, plus a second one: the repeated <value> case turned out to arrive as an array from fast-xml-parser, so the actual collision branch (two different prefixes bound to the same URI, or a prefix and an inline xmlns) still went untested. Both are covered now.

dschmidt:
Pushed 6e09d61. Beyond the three comments, digging into the first one turned up two more things in my own code that I fixed in the same commit:

- the retained xmlns declarations leaked into the result as `@xmlns*` attributes. Worst case was an empty multistatus: the leftover object slipped past the empty check in `normaliseResult` and produced a phantom response entry. They are dropped now once resolved, and an element left empty by that collapses to `""` like in the default mode.
- a namespace declared on the property element itself with a prefix (`<x:custom xmlns:x="...">`) was silently falling back to the null namespace. Now folded into the scope before resolving.

One limitation I left in place deliberately, and pinned with a test so it is explicit: only the direct children of `<prop>` are rewritten, so nested elements keep the prefix the server used (`{DAV:}resourcetype` comes back as `{ "d:collection": "" }`). Recursing all the way down is doable, happy to do it if you would rather have that.

perry-mitchell:
Thanks @dschmidt! Appreciate the quick response. I think it's fine as is, and the recursion could be added later if needed. I'll take this as it's quite tidy now, aiming for a release later today.

dschmidt:
Amaaazing! Thanks a bunch, much appreciated :)

perry-mitchell:
Released under 5.11.0

dschmidt:
Thanks a lot! 

Did you see the followup at https://github.com/perry-mitchell/webdav-client/pull/415 tho? 🫣

perry-mitchell:
I did briefly but haven’t had time to check it yet, thought it best to release as it had other items in it. I’ll try to check it asap. Thanks!

PR #411, opened 2026-06-24
fix: declare @nodable/entities as a dependency

JammingBen:
`tools/dav.ts` imports `EntityDecoder` from `@nodable/entities`, but the package was never declared in dependencies. Under strict `node_modules` layouts (e.g. pnpm) the import fails to resolve with:

```
Cannot find package '@nodable/entities' imported from webdav/dist/node/tools/dav.js
```

Hence declare it explicitly (`^2.2.0`, matching `fast-xml-parser`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@buttercup/fetch": "^0.2.1",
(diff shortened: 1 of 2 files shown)


perry-mitchell:
Thanks @JammingBen !

perry-mitchell:
Released under 5.11.0

